PR #1, opened 2022-01-12
Let's start this project off!

rob-baron:
This is the initial branch of for a development environment to be used for integrating the script(s) needed to pull data from OpenStack and bring it into xdmod.

currently there are 2 out of date scripts that are used, but I suspect there are more needed to complete this.


rob-baron:
don't worry, this one is going away as:

1) it contains my thoughts on how to do it an how I proceeded to work on this project
2) Once it became a draft, no one commented or suggested directions within the draft.

PR #14, opened 2022-03-18
type analysis of xdmod-openstack specific structure

rob-baron:
just listing the event types with different xdmod-openstack structures


rob-baron:
Am closing this PR as no one has reviewed it and I placed most of it in issues in the backlog/sprint.

PR #20, opened 2022-04-04
Xdmod k8s deployment

rob-baron:
Need to write up coherent instruction on how it was installed

Need to write some scripts for the init and setup deployments.  And yes, I think I agree with lars that init containers will do.


Dockerfile.centos7:xdmod-base
```diff
@@ -0,0 +1,58 @@
+FROM centos:7.9.2009
```
larsks, line 1: You mentioned that this was copied from the xdmod project, but I'm not sure that we need it. Installing the `xdmod` rpm itself will bring in any necessary dependencies, so the other Dockerfiles should be fine if they were simply based on a recent CentOS image instead.

Dockerfile.xdmod
```diff
@@ -0,0 +1,51 @@
...
+WORKDIR /root
+
+
+# This installs a barebones 10.0 instance of XDmod
+RUN mkdir -p /root/rpmbuild/RPMS/noarch \
+    && mkdir /tmp/supremm \
+    && yum -y install openssh-server openssh-clients pcp pcp-manager pcp-conf pcp-libs python-pcp perl-PCP-PMDA pcp-system-tools pcp-pmda-gpfs pcp-pmda-lustre pcp-pmda-infiniband pcp-pmda-mic pcp-pmda-nvidia-gpu \
+    pcp-pmda-nfsclient pcp-pmda-perfevent pcp-pmda-json python3 \
+    && pip3 install --upgrade pip \
+    && pip3 install keystoneauth python-keystoneclient python-novaclient python-cinderclient cryptography python-openstacksdk mysql-connector-python
```
larsks, line 16: The bulk of this block (setting up an `rpmbuild` directory, installing an `ssh` server, and installing all the `pcp` related packages) is unnecessary. Probably we don't need the `ssh` client, either. It looks like really we just need to install `python` and then continue with the `pip` commands.

Dockerfile.xdmod
```diff
@@ -0,0 +1,51 @@
...
+
+# RPM install of XDMod
+RUN cd /root/rpmbuild/RPMS/noarch \
+    && wget -nv https://github.com/ubccr/xdmod/releases/download/v10.0.0/xdmod-10.0.0-1.0.el7.noarch.rpm \
+    && yum -y install xdmod-10.0.0-1.0.el7.noarch.rpm
```
larsks, line 21: These three lines could be replaced by:

```
RUN yum -y install https://github.com/ubccr/xdmod/releases/download/v10.0.0/xdmod-10.0.0-1.0.el7.noarch.rpm
```

Dockerfile.xdmod
```diff
@@ -0,0 +1,51 @@
...
+
+COPY ./moc_xdmod_conf/httpd.conf /etc/httpd/conf/httpd.conf
```
larsks, line 23: This file doesn't exist in the repository.

Dockerfile.xdmod
```diff
@@ -0,0 +1,51 @@
...
+# am copying the xdmod-httpd conf file to ssl.conf as both use *:443
+# and this is easier than just deleting the ssl.conf
```
larsks, line 25: It looks like you're just deleting `ssl.conf`.

Dockerfile.xdmod
```diff
@@ -0,0 +1,51 @@
...
+RUN rm -f /etc/httpd/conf.d/ssl.conf \
+    && rm -f /run/httpd/httpd.pid
+
+COPY ./test_data/openstack.2018-04-17T00-00-00_2018-04-30T23-59-59.json /root/test/openstack/2018-04-17T00-00-00_2018-04-30T23-59-59.json
```
larsks, line 29: I wonder if test data should just be mounted in a runtime rather than baked into the image?

Dockerfile.xdmod
```diff
@@ -0,0 +1,51 @@
...
+
+COPY ./hypervisor_facts.py /usr/bin/xdmod-openstack-hypervisor
+COPY ./moc_openstack_api_reporting.py /usr/bin/xdmod-openstack-reporting
+RUN chmod 777 /usr/bin/xdmod-openstack-hypervisor \
+    && sed -i -e 's/\r$//' /usr/bin/xdmod-openstack-hypervisor \
```
larsks, line 34: It looks like you're doing some end-of-line conversion here. That would make sense if we were pulling a file from some third-party source, but since it lives right here in the repository why not just ensure it's using correct end-of-line conventions in the repository and remove these `sed` statements?

Dockerfile.xdmod
```diff
@@ -0,0 +1,51 @@
...
+COPY ./moc_xdmod_conf/httpd.conf /etc/httpd/conf/httpd.conf
+# am copying the xdmod-httpd conf file to ssl.conf as both use *:443
+# and this is easier than just deleting the ssl.conf
+RUN rm -f /etc/httpd/conf.d/ssl.conf \
+    && rm -f /run/httpd/httpd.pid
+
+COPY ./test_data/openstack.2018-04-17T00-00-00_2018-04-30T23-59-59.json /root/test/openstack/2018-04-17T00-00-00_2018-04-30T23-59-59.json
+
+COPY ./hypervisor_facts.py /usr/bin/xdmod-openstack-hypervisor
+COPY ./moc_openstack_api_reporting.py /usr/bin/xdmod-openstack-reporting
```
larsks, line 32: These files don't exist in the repository.

Dockerfile.xdmod
```diff
@@ -0,0 +1,51 @@
...
+
+EXPOSE 8080
+
+# this is done so to log into the container and manually run xdmod-setup
+#ENTRYPOINT ["systemctl", "start", "httpd.serivce"]
+#ENTRYPOINT ["apachectl", "start"]
+#ENTRYPOINT ["sleep", "1000000"]
+
+CMD ["-D", "FOREGROUND"]
+ENTRYPOINT ["/usr/sbin/httpd"]
```
larsks, line 49: I think it would make more sense to just have:

```
CMD ["/usr/sbin/httpd", "-D", "FOREGROUND"]
```

There's no reason to set the `ENTRYPOINT` here, and it makes it more difficult if someone just wanted to spawn a shell instead of running Apache.

Dockerfile.xdmod.init
```diff
@@ -0,0 +1,46 @@
+FROM moc-xdmod-base
```
larsks, line 1: This seems to be largely identical to `Dockerfile.xdmod`. What's the difference between the resulting two images?

Dockerfile.xdmod.init
```diff
@@ -0,0 +1,46 @@
...
+
+#COPY assets /tmp/assets
+#RUN /tmp/assets/copy-caches.sh
+#COPY bin /root/bin
+
+WORKDIR /root
+
+
+# This installs a barebones 9.5 instance of XDmod
```
larsks, line 10: This comment doesn't match version of xdmod actually installed by the Dockerfile.

assets/mysql-server.cnf
```diff
@@ -0,0 +1,15 @@
+[mysqld]
```
larsks, line 1: This is a configuration file for mysql, but we're never running mysql/mariadb inside the xdmod containers. The database service in the k8s configuration in this repository is provided by the upstream mariadb image.

assets/vimrc
```diff
@@ -0,0 +1,86 @@
+if v:lang =~ "utf8$" || v:lang =~ "UTF-8$"
```
larsks, line 1: I don't think we need to install a vim configuration into our Docker images. In general we shouldn't be editing things inside the image.

xdmod-pv.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    type: local
+spec:
+  storageClassName: manual
+  capacity:
+    storage: 5Gi
+  accessModes:
+    - ReadWriteOnce
+  Reclaim Policy: Retain
+  hostPath:
```
larsks, line 14: We generally won't use `hostPath` volumes in our OpenShift deployments. They require elevated privileges, and they can't be accessed from different nodes like a regular PVC.

Dockerfile.centos7:xdmod-base
rob-baron: At this stage the xdmod-base loads tooling that is convenient (vi, wget ... ) that is use in building xdmod.

Dockerfile.xdmod
rob-baron: done.

xdmod-pv.yaml
rob-baron: This is how I am creating the PV on minikube.  As I don't need to worry about creating PVs on OpenShift.

Would it be better for me to keep the minicab deployment locally, or should it be under source control?

Dockerfile.centos7:xdmod-base
larsks: We're not actually building xdmod, right? We're just consuming the RPM. But if it's convenient to get things going might as well keep it for now; we can also revisit later.

Dockerfile.xdmod.init
larsks: I think a single image would be best (and then just override the command as necessary in your deployment manifests).

xdmod-pv.yaml
larsks: I didn't realize you were explicitly targeting minikube right now. If that's the case, I guess leave these here for now and we'll revisit when we get around to deploying in openshift.

rob-baron: I am planning on both.  minikube is more convenient for development.

I am keeping it separated, that the deployment to openshift under /k8s/..., and the pv definitions specifically for minikube out of that directory for now.

Dockerfile.xdmod
rob-baron: done

Dockerfile.xdmod
rob-baron: done

Dockerfile.xdmod.init
rob-baron: I am now creating a single docker file.  In some regards it should be teased apart into 3 different docker files:

1.  to copy the configuration files and source code to their respective directories
2. to run the xdmod-setup script (doesn't really need the web server, as an example)
3. to run xdmid's UI

but that is for a different day

Dockerfile.xdmod.init
rob-baron: done

assets/mysql-server.cnf
rob-baron: The xdmod folks are running mariadb out of the same container that xdmod is running.  The example that I am working from used a separate container to run mariadb.  I went with running mariadb in a separate container as that made more sense, though it is a bit more difficult to setup.

Dockerfile.xdmod
rob-baron: done.

assets/vimrc
rob-baron: I find it useful for debugging things within the container.  I will not have in the production container - however nothing seems to be dependent on vim, so it is easily removed.

rob-baron:
Since there is quite a bit to review, might as well start.  I've addressed most of Larsks' original comments, though I plan on doing more pruning of the docker build file, replacing the route with an ingress - and I am sure I'm forgetting some other things to try.

Dockerfile.xdmod
rob-baron: I have eliminate pcp.  However, xdmod seems to require the openssl-server and openssl-clients with the way that they are doing things.  I have simplified this a bit.

Dockerfile.xdmod.init
rob-baron: I eliminated the Dockerfile.xdmod.init.

assets/vimrc
rob-baron: Ok, I have removed vi from the "production" container and moved it to a "dev" container based off of the production container.

the build configs/image streams have yaml for them.

Dockerfile.xdmod
rob-baron: I'll move it to the "dev" build.

assets/mysql-server.cnf
rob-baron: Actually assets/mysql-server.cnf and the vi configuration files are not being used, though they may be useful to include in the development container.

Dockerfile.xdmod
rob-baron: I just kept this as it was the easiest way to get something working.  will look at removing this in the future.

Dockerfile.centos7:xdmod-base
rob-baron: The base RPM doesn't include all of its dependencies.

Dockerfile.xdmod
rob-baron: Actually, I can eliminate this as I can copy in the test data with oc rsync.

joachimweyl:
I don't see any open conversations about changes that need to be made. What is the next step for this PR?

Dockerfile.moc-xdmod
```diff
@@ -0,0 +1,100 @@
...
+    && sed -i -e 's/\r$//' /usr/bin/xdmod-openstack-reporting \ 
+    && chmod 777 /usr/bin/xdmod-init \
+    && sed -i -e 's/\r$//' /usr/bin/xdmod-init \ 
+    && chmod -R 777 /usr/bin \
+    && chmod -R 777 /var/log \
+    && chmod -R 777 /var/lib \
+    && chmod -R 777 /root \
+    && chmod -R 777 /run/httpd \
+    && chmod -R 777 /etc/xdmod \
+    && chmod -R 777 /etc/httpd
```
larsks, line 94: It really shouldn't be necessary to open up access like this on so many directories. We ought to be able to solve these with more granular permissions, and possibly the correct use of volumes for things that need to be writable like `/var/log` and `/run` (the latter of which could be a `tmpfs` mount).

joachimweyl:
Rob - "It is basically done, I have some more testing to do today with respect to the generalizations (which I had to rework over the weekend).  Both, with a couple of the dependencies, disappearing, and realizing the way that I had laid it out wasn’t the best way forward."

joachimweyl:
@rob-baron it appears as though all comments are resolved or almost there. What are the next steps?

Dockerfile.moc-xdmod
```diff
@@ -0,0 +1,95 @@
...
+    && chmod 777 /usr/bin/xdmod-openstack-reporting \
+    && sed -i -e 's/\r$//' /usr/bin/xdmod-openstack-reporting \ 
+    && chmod 777 /usr/bin/xdmod-init \
+    && sed -i -e 's/\r$//' /usr/bin/xdmod-init \ 
+    && chmod -R 777 /usr/bin \
+    && chmod -R 777 /var/log \
+    && chmod -R 777 /var/lib \
+    && chmod -R 777 /root \
+    && chmod -R 777 /run/httpd \
+    && chmod -R 777 /etc/xdmod
```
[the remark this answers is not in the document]
rob-baron: Multiple processes need to read and write to the /root directory as that is where I am mounting xdmod data directory

/run/httpd permissions were too restrictive for it to be started by the container

/etc/xdmod needs to be modified by xdmod-setup 

/var/lib needs read write permissions for xdmod's php script stores information there (I seem to remember a pid file and/or cookies).

These are the directories that xdmod/php uses by default, I cannot recommend we put them all under /var/log.  Furthermore, I cannot recommend mounting volumes at run time as we don't need to persist that state information between runs.

README.md
```diff
@@ -1,2 +1,100 @@
...
+5)  create the necessary configmaps/secrets
+     
+    kubectl create cm --from-file xdmod_conf/xdmod_init.json cm-xdmod-init-json 
+    oc create cm --from-file xdmod_conf/xdmod_init.json cm-xdmod-init-json 
+
+    kubectl create cm --from-file xdmod_conf/httpd.conf cm-httpd-conf
+    oc create cm --from-file xdmod_conf/httpd.conf cm-httpd-conf
+
+
+6) Create the namespace (project) - use kubectl for minikube, or oc for openshift
```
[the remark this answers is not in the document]
rob-baron: I'll change the instruction.

rob-baron:
kim,

My next steps are:

1) finish addressing the comments.
2) get the install on cop-staging working.
3) redo the install on the nerc

hypervisor_facts.py
```diff
@@ -0,0 +1,128 @@
...
+def get_latest_facts(config):
+    """gets the previously recorded hypervisor facts"""
+    file_list = glob.glob(config["outdir"] + "//hypervisor_facts_*.json")
+
+    if len(file_list) == 0:
+        return {"hypervisors": []}
+
+    newest = max(file_list, key=lambda d: datetime.datetime.strptime(d, "{}/hypervisor_facts_%Y-%m-%dT%H:%M:%S.json".format(config["outdir"])))
+
+    f = open(newest, "r")
```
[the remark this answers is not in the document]
rob-baron: done.

joachimweyl:
@rob-baron  based on your standup notes it sounds like 2 and 3 are done and if your flurry of activity this morning is any indication you are well on your way to addressing all of the comments. Well done.

Dockerfile.moc-xdmod
rob-baron: I was able to change the /run/httpd to just that directory.

Keep in mind, we are running this as a user process and not as root (as is expected by the xdmod team).   So here is a run down of the all of the permissions that you are mentioning:

        chmod -R 777 /usr/bin      # stores the xdmod executables (xdmod-setup, xdmid-shredder ... )
        chmod 777 /usr/bin         # This gave the error that xdmod-setup could not be run

        chmod -R 777 /var/log      # this needs to store log files written by a user process of apache, php and xdmod
        chmod -R 777 /var/lib      # this is used to store the session cookies used by xdmod

        chmod -R 777 /root         # This stores the data
        default                    # unable to create the subdirectories for the data        

        chmod 777 /run/httpd       # This is the directory of the pid file (removed the -R) 

        chmod -R 777 /etc/xdmod     # stores the configuration of xdmod
        default                     #  Was able to copy, but not able to modify  

        chmod -R 777 /etc/httpd  # stores all of the configuration for the apache    
        default                     # Was able to copy, but not able to modify

What permissions would be restrictive enough for you?  Are there a particular set you want me to test?

Dockerfile.moc-xdmod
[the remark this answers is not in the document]
rob-baron: I'm marking this as resolved as it is it duplicated below.

Dockerfile.moc-xdmod
rob-baron: @larsks - do you have an opinion on this?

larsks: Yes, and sorry for the slow reply.

Looking at the deployment, we're mounting the following volumes:

```
- name: vol-xdmod-conf
  mountPath: /etc/xdmod
- name: vol-xdmod-src
  mountPath: /usr/share/xdmod
- name: vol-xdmod-data
  mountPath: /root/xdmod_data
- name: vol-httpd-conf
  mountPath: /etc/httpd
- name: vol-clouds-yaml
  mountPath: /etc/openstack
- name: vol-httpd-conf
  mountPath: /root/httpd
```

The user under which the container is running will already have
write access to all of these directories, although you will not be
able to access directories under `root`. We can demonstrate both facts
with the following example:

```
apiVersion: apps/v1

kind: Deployment
metadata:
  name: permission-test
spec:
  template:
    spec:
      containers:
        - image: docker.io/alpine:latest
          name: permission-test
          volumeMounts:
            - name: vol-xdmod-conf
              mountPath: /etc/xdmod
            - name: vol-xdmod-src
              mountPath: /usr/share/xdmod
            - name: vol-xdmod-data
              mountPath: /root/xdmod_data
          command:
            - sh
            - -c
            - |
              set -e

              for dir in /etc/xdmod /usr/share/xdmod /root/xdmod_data; do
                cd $dir
                pwd
                date > testfile
              done

              sleep inf
      volumes:
        - name: vol-xdmod-conf
          persistentVolumeClaim:
            claimName: pvc-xdmod-conf
        - name: vol-xdmod-src
          persistentVolumeClaim:
            claimName: pvc-xdmod-src
        - name: vol-xdmod-data
          persistentVolumeClaim:
            claimName: pvc-xdmod-data
```

The pod from that deployment will output:

```
/etc/xdmod
/usr/share/xdmod
sh: cd: line 3: can't cd to /root/xdmod_data: Permission denied
```

That is, it demonstrably is able to write files to `/etc/xdmod` and
`/usr/share/xdmod` without making any permissions changes. It is
unable to access `/root/xdmod_data` because the permissions on the
`/root` directory are:

```
$ ls -ld /root
drwx------    1 root     root            24 Jun 10 16:04 /root
```

But that doesn't mean we should change the permissions on `/root`:
because we *know* we're going to be running as a non-`root` user, we
should be using directories in a different location (which could be
pretty much anything: `/data`, `/home/xdmod`, whatever you want).

For directories like `/var/log/xdmod` and `/var/lib/<whatever>`, a
common solution would be to mount an `emptyDir` volume at both those
locations.  E.g., if we were to modify the above example `Deployment`,
we would add the following to the `volumes` section:

```
      volumes:
        - name: xdmod-state
          emptyDir: {}
```


And then the following to the `volumeMounts` section in our
containers:

```
          volumeMounts:
            - name: xdmod-state
              mountPath: /var/log/xdmod
            - name: xdmod-state
              mountPath: /var/lib/xdmod
```

With that volume mounted, while we cannot write to `/var/log` or
`/var/lib`:

```
~ $ touch /var/log/testfile
touch: /var/log/testfile: Permission denied
~ $ touch /var/lib/testfile
touch: /var/lib/testfile: Permission denied
```

We can write to `/var/log/xdmod` and `/var/lib/xdmod` without a
problem:

```
~ $ touch /var/log/xdmod/testfile
~ $ touch /var/lib/xdmod/testfile
```

I would not expect anything to be *writing* to `/usr/bin` at runtime,
so I'm not sure what's going on there.

larsks: ...and since you're also mounting a volume on `/etc/httpd`, that directory should also be writable by the current user without any permission changes.

rob-baron: Do we really need a mount a separate disk to hold the pid file as we already have ephemeral storage (as in the container once it is deployed)?

joachimweyl:
@rob-baron it looks like the comments have been almost all dealt with. Do we have any last pieces that we need to change?

rob-baron:
larsks, I'll time box this effort and merge this PR this evening, either with changes that support what you want or without and add an issue.

joachimweyl:
> larsks, I'll time box this effort and merge this PR this evening, either with changes that support what you want or without and add an issue.

@rob-baron just to confirm when you said you would merge the PR with or without the changes. I assume by that you mean if there are no changes you will merge and if there are changes, the changes will require review and therefore it will not be merged yet?

rob-baron:
ok, I'll merge it at the last review state.

knikolla (approved):

PR #30, opened 2022-05-13
Xdmod openstack reporting

rob-baron:
This is to create the xdmod-openstack-reporting script to pull event data from openstack without using ceilometer or panko as both of those openstack components have been either removed or deprecated.


moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+            vm_timestamps[server_id]["updated"] = 1
+
+    json_out = f"{config['outdir']}/{last_run_timestamp}_{script_timestamp}.json"
+    with open(json_out, "w+", encoding="utf-8") as outfile:
+        json.dump(events, outfile, indent=2, sort_keys=True, separators=(",", ": "))
+    print(f"output file: {json_out}")
+
+    for key, val in vm_timestamps.items():
+        if val["updated"] == 0:
+            del vm_timestamps[key]
```
naved001, line 559: doesn't this produce a runtime error in python3 when a dictionary is modified while you are iterating over it?

tmp.md
```diff
@@ -0,0 +1,1507 @@
+
```
naved001, line 1: is this file meant to be formatted in some way that makes it easier to read?

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+#     "service": "compute",
+#     "state": "active",
+#     "tenant_id": "4aeb007a4f9020333a1a1be224bef276",
+#     "user_id": "3abcb51ff942a45b52ac90915ef4c7fb",
+#     "user_name": "yerwa",
+#     "vcpus": "1",
+# }
+
+
+def convt_to_ceilometer_event_types(event):
```
naved001, line 324: you might as well name it `convert_to_....`. I don't think saving 2 characters makes it any longer

tmp.md
naved001: My comment was about the whole file and not a specific line, but I had to pick a link to make a comment.

Anyway, looks like it's a `|` separated file. Since it's called `tmp.md` I was thinking it's markdown. Maybe don't call it `*.md` then.

After opening it in excel and setting the custom delimiter to `|` it made sense.

rob-baron: I was planning to use markdown, so I went with ‘|' as the separator.  Unfortuantely, neither the markdown plugin or the csv plugin that I tried for visual studio code worked well, so I just went with using excel.

In general I hate to use excel, however, I had already spent a day to produce this and had gotten the information that I needed from this file.  It is just here for documenting things and nothing else.

moc_openstack_api_reporting.py
naved001:
```
>>> vm_timestamps = {"A": 1, "B": 2, "C": 3, "D": 4}
>>> for key, val in vm_timestamps.items():
...     if val==2:
...             del vm_timestamps[key]
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
```

Why can't I do this then?

tmp.md
naved001: > I was planning to use markdown, so I went with ‘|' as the separator. 

You need `----` to separate the headers in addition to '|' in a markdown table. https://www.markdownguide.org/extended-syntax/#tables

> It is just here for documenting things and nothing else.

If something is going to be added into a repository then it's purpose should be clear at least.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+    for project in openstack_conn.list_projects():
+        project_dict[project.id] = project
+
+    volume_dict = {}
+    for volume in openstack_cinder.volumes.list(search_opts={"all_tenants": True}, detailed=True):
+        volume_dict[volume.id] = volume
+
+    events = []
+    server_dict = {}
+    ap_t = datetime.datetime.strptime(script_timestamp, date_format_str)
```
naved001, line 515: what's `ap_t`?

moc_openstack_api_reporting.py
rob-baron: I was not getting that runtime error.  will change this.

moc_openstack_api_reporting.py
rob-baron: audit_period_time - honestly it isn't actually the audit_period_time, but the first timestamp in the series of events.  technically, the audit period shouldn't be based on the first timestamp in the series of events, it should be based on either an audit period for a grant, or an audit period for xdmod.

Have since found that xdmod is not using audit period for anything yet.  I have found series of events in the reference file that all events for a vm fall within a single audit period, and sometimes they are split into 2 audit periods.  So I can get away with setting it for the minimum event time to the current time of the script.  Note, that this is only done when the script runs for the first time.

moc_openstack_api_reporting.py
naved001: > I was not getting that runtime error

maybe it didn't hit the branch where the delete takes place.

moc_openstack_api_reporting.py
naved001: okay, I think having a more descriptive name would be better.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+    """The main function"""
+    config = {}
+
+    do_parse_args(config)
+    do_read_config(config)
+
+    openstack_conn = openstack.connect(cloud=config["cloud"])
+    openstack_nova = nova_client.Client(2, session=openstack_conn.session)
+    openstack_cinder = cinder_client.Client(3, session=openstack_conn.session)
+    openstack_keystone = keystone_client.Client(3, session=openstack_conn.session)
```
larsks, line 476: This variable isn't used anywhere.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+    for volume in openstack_cinder.volumes.list(search_opts={"all_tenants": True}, detailed=True):
+        volume_dict[volume.id] = volume
+
+    events = []
+    server_dict = {}
+    ap_t = datetime.datetime.strptime(script_timestamp, date_format_str)
+    for server in openstack_nova.servers.list(search_opts={"all_tenants": True}, detailed=True):
+        server_dict[server.id] = compile_server_state(server, project_dict, flavor_dict, user)
+        event_data = {"event_type": "compute.instance.exists", "event_time": script_timestamp}
+        if server_t < ap_t:
```
larsks, line 519: I don't think `server_t` is defined anywhere (prior to this reference).

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
+#!/usr/bin/python3
+""" This is intended to poll openstack for changes to be reported in xdmod """
+# pylint: disable=line-too-long too-many-locals
```
larsks, line 3: I don't think you need to disable the `too-many-locals` check.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+    parser.add_argument("-s", "--start", help="Start time for records", required=False)  # pulls the time of the event from the event list
+    parser.add_argument("-e", "--end", help="End time for records", required=False)  # defaults to current time
+    parser.add_argument("-o", "--outdir", help="Output directory")
+    parser.add_argument("--cloud", help="Using the name/credentials in the cloud config file to connect with OpenStack")
+    parser.add_argument("-d", "--db", help="Database name, only valid for --use-db")
+    parser.add_argument("-f", "--filter-noise", help="Filter out noisy events. Requires panko patch", action="store_true")
+
+    args = parser.parse_args()
+
+    config["loglevel"] = logging.CRITICAL
```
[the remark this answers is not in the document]
rob-baron: I kept this the same as the script that xdmod folks are using in order to keep their interface the same.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+        config["collapse_traits"] = True
+
+    if args.filter_noise:
+        config["filter_noise"] = True
+
+    if args.cloud:
+        config["cloud"] = args.cloud
+
+    config["config_dir"] = "."
+    if args.config_dir:
```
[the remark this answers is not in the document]
rob-baron: I kept this the same as the script that xdmod folks are using in order to keep their interface the same.

moc_openstack_api_reporting.py
rob-baron: Yes, I used it at one point, but then didn't really need it.  Just didn't get around to eliminating it here.  Will do so.

moc_openstack_api_reporting.py
rob-baron: I must have refactored the code so that this is no longer necessary.  This is the basic problem I have with disabling lint errors at the file scope, instead of, say the class or the function scope.

moc_openstack_api_reporting.py
rob-baron: I would investigate it further, but it is a waste of time, easier to pull the array of keys, then iterate though them to delete the ones that are no-longer needed.

I have already re-written it this way last friday, and haven't gotten around to testing it fully.

tmp.md
rob-baron: ok, I'll delete this from the repo

moc_openstack_api_reporting.py
rob-baron: So, how to put this, I am trying to get this to produce data.  I have renamed ap_t to something more meaningful.  In doing so, realized that this does need to be refactored to be easier to read.

So I am trying to balance the ease of reading with the urgency to produce data.

moc_openstack_api_reporting.py
naved001: > I have already re-written it this way last friday, and haven't gotten around to testing it fully.

will you commit the fix in this PR? I think if we iterate over `list(vm_timestamps.items())` then that should take care of things.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+        "instance_id": vm.id,  # ID for the VM instance,
+        "record_type": "",  # Type of record from list in modw_cloud.record_type table,
+        "block_devices": [],
+        "private_ip": private_ip,  # "Private IP address used by the instance,
+        "root_type": "",  # "Type of storage initial storage volume is, either ebs or instance-store
+    }
+
+    vm_volumes = []
+    if "os-extended-volumes:volumes_attached" in vm.to_dict():
+        vm_volumes = vm.__getattribute__("os-extended-volumes:volumes_attached")
```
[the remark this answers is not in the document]
larsks: @knikolla do you have an opinion here? I don't feel strongly about it, because obviously what's here does work.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+    server_event = copy.deepcopy(server_state)
+    if event["event_type"] == "compute.instance.exists":
+        server_event["audit_period_beginning"] = copy.copy(event["audit_period_start"])
+        server_event["audit_period_ending"] = copy.copy(event["audit_period_end"])
+    server_event["event_type"] = copy.copy(event["event_type"])
+    server_event["generated"] = copy.copy(event["event_time"])
+    return server_event
+
+
+def get_session_for_resource(resource):
```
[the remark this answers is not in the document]
rob-baron: This was shamelessly copied from kristi, as this is how cold front connects to OpenStack.  Personally, I find keeping all openstack credentials in a clouds.yaml file as it is easier to call the program with "-cloud <openstack_name>"  as opposed to setting up the 3 environment variables per cloud, and using them within the the program - I also coded it this way from the beginning).

It has been eliminated.

moc_openstack_api_reporting.py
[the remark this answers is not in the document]
rob-baron: I just went with getattr.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,586 @@
...
+    server_dict = {}
+    min_event_time = script_datetime
+    for server in openstack_nova.servers.list(search_opts={"all_tenants": True}, detailed=True):
+        server_dict[server.id] = compile_server_state(server, project_dict, flavor_dict, user)
+        server_t = datetime.datetime.strptime(server.created, "%Y-%m-%dT%H:%M:%SZ")
+
+        if server_t < min_event_time:
+            min_event_time = server_t
+        if server.id in vm_timestamps:
+            event_data["start_ts"] = vm_timestamps[server.id]
```
larsks, line 507: `event_data` isn't defined at this point in the code.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,586 @@
...
+import logging
+import datetime
+import re
+import openstack
+import copy
+import os
+from novaclient import client as nova_client
+from cinderclient import client as cinder_client
+from keystoneclient import client as keystone_client
+import pprint
```
larsks, line 16: There are a bunch of unused imports here:

```
moc_openstack_api_reporting.py:9:1: F401 're' imported but unused
moc_openstack_api_reporting.py:12:1: F401 'os' imported but unused
moc_openstack_api_reporting.py:15:1: F401 'keystoneclient.client as keystone_client' imported but unused
moc_openstack_api_reporting.py:16:1: F401 'pprint' imported but unused
```

Not critical, but worth fixing if you're making other changes.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,586 @@
...
+    """reads from the config file if present"""
+    try:
+        with open(f"{config['config_dir']}/openstack_reporting.json", "r", encoding="utf-8") as config_fp:
+            newconfig = json.load(config_fp)
+            config.update(newconfig)
+    except IOError:
+        return
+    except ValueError:
+        print("Error, Badly formatted config file")
+        exit()
```
larsks, line 196: Consider replacing these two lines with:

```
exit('Error, badly formatted config file')
```

This ensures that the error is printed on *stderr* rather than *stdout*.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,586 @@
...
+            new_event["generated"] = (time_0 + datetime.timedelta(seconds=delta_time)).isoformat()
+            ret_list.append(new_event)
+            delta_time += 10
+        return ret_list
+    return []
+
+
+def compile_server_state(server, project_dict, flavor_dict, user_dict):
+    """This one is implied.  We can tell this from the state of the VM"""
+    launched_ts = None
```
larsks, line 385: It doesn't make any sense to set `launched_ts = None` here, since you're setting it unconditionally to another value on the following line (and similar for the `host` and `terminated_ts` variables, below).

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,586 @@
...
+        "block_devices": [],
+        "private_ip": private_ip,  # "Private IP address used by the instance,
+        "root_type": "",  # "Type of storage initial storage volume is, either ebs or instance-store
+    }
+
+    vm_volumes = []
+    vm_volumes = getattr(vm, "os-extended-volumes:volumes_attached", [])
+    for volume_id in vm_volumes:
+        vol = volume[volume_id["id"]]
+        vol_project_id = ""
```
larsks, line 234: You don't need to initialize `vol_project_id` or `backing` here.

larsks (changes requested):
There is one actual bug (an undefined variable; I've marked the comment with a :rocket: since that seemed the best option of the available emojis). The remaining comments are just suggestions; I think they're worth changing as long as you're making changes, but they don't impact the behavior of the code.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,586 @@
...
+        "resource_id": server.id,
+        "root_gb": str(flavor_dict[server.flavor["id"]]["disk"]),
+        "service": "compute",
+        "state": server.status,
+        "tenant_id": server.tenant_id,
+        "user_id": server.user_id,
+        "user_name": user_name,
+        "vcpus": str(flavor_dict[server.flavor["id"]]["vcpus"]),
+    }
+    if terminated_ts is not None:
```
naved001, line 433: At line 390:

`terminated_ts = getattr(server, "OS-SRV-USG:terminated_at", "")`

`terminated_ts` would either bet set to some value or be set to the default `""` which is not `None`. So it looks like we will always hit this branch. Unless of course the attribute could actually return `None`. Am I reading this correctly?

moc_openstack_api_reporting.py
rob-baron: yes, I didn't do the refactoring using getattr correctly.

moc_openstack_api_reporting.py
rob-baron: I've removed them despite the utility that I occasionally find in just keeping them around when I am developing things.

rob-baron:
@joachimweyl - yes they are, however, I am more worried with getting data in the format that xdmod will accept, of which, even though this has produced output (not necessarily this particular version with getattr?), I have modified it with getattr as per larsks suggestion and saved an intermediate form as my os wanted to update itself.

I will check in the next update when I can test again.

moc_openstack_api_reporting.py
rob-baron: yes, naved, I didn't refactor using getattr correctly.

joachimweyl:
@rob-baron is this blocked until we hear back from XDMoD about #33 ?

joachimweyl:
@rob-baron with the suggestions that XDMoD provided is this no longer blocked?

rob-baron:
that is correct.

moc_openstack_api_reporting.py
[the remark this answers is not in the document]
rob-baron: fine larsks, if it will satisfy your design sensibilities I'll remove this as I am only using the --cloud command line option.

moc_openstack_api_reporting.py
```diff
@@ -0,0 +1,572 @@
...
+        "compute.instance.power_off": ["start", "end"],
+        "compute.instance.power_on": ["start", "end"],
+        "compute.instance.shutdown": ["start", "end"],
+    }
+    if event["event_type"] == "compute.instance.create" and event["state"] == "ERROR":
+        event["event_type"] = "compute.instance.create.error"
+        return [event]
+    if event["event_type"] in map_current_to_old:
+        event["event_type"] = map_current_to_old[event["event_type"]]
+    if event["event_type"] in ceilometer_event_types:
```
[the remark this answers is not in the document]
rob-baron: So larsks, help me understand what you need so that I know either how I should respond or what needs to happen to resolve this conversation.

joachimweyl:
@rob-baron is the `moc_openstack_api_reporting.py` conflict a relatively easy fix or is it going to require a large chunk of time?

moc_openstack_api_reporting.py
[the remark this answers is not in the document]
rob-baron: I am going to resolve this as I have not heard back from you concerning this conversation

rob-baron:
@larsks, I have made the change that you requested.  It will change the behavior as compared with the process that the xdmod team is developing, but not in a significant manner.

larsks (approved):

rob-baron:
I will just go with larks' approval as kristi is probably busy.

PR #39, opened 2022-06-16
Allows phpssid to be set even if it is not an https connection

rob-baron:
currently, I am unable to log in due to the web browser not storing cookies (specifically, the phpssid token).  Before I turn on SSL at the route, I want to ensure that xdmod is configured to correctly.

### DIFF
--- a/xdmod_init.py
+++ b/xdmod_init.py
@@ -371,47 +371,6 @@ def main():
             print(" Organization ")
             xdmod_setup_organization(xdmod_init_json["organization"])
 
-        print("   Alowing admin and accounts to log into xdmod without ssl")
-        # replace:
-        #    setcookie('xdmod_token', getToken(), 0, '/', '', true, true);
-        # with:
-        #    setcookie('xdmod_token', getToken\(\), 0, '/', '', false, false)'
-        if os.path.isfile("/usr/share/xdmod/libraries/rest.php"):
-            with open("/usr/share/xdmod/libraries/rest.php", encoding="utf-8") as file:
-                rest_php = file.read().replace(
-                    "setcookie('xdmod_token', getToken(), 0, '/', '', true, true)",
-                    "setcookie('xdmod_token', getToken(), 0, '/', '', false, false)",
-                )
-            with open("/usr/share/xdmod/libraries/rest.php", "w", encoding="utf-8") as file:
-                file.write(rest_php)
-
-        if os.path.isfile("/usr/share/xdmod/libraries/security.php"):
-            """
-                  $cParams["lifetime"],
-                  $cParams["path"],
-                  $cParams['domain'],
-            -     true
-            +     false
-                  );
-            """
-            security_php = None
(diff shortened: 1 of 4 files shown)


xdmod_init.py
```diff
@@ -384,6 +385,33 @@ def main():
...
+                    security_php,
+                )
+                if match1_str:
+                    match2_str = re.search(r"true", match1_str.group())
+                    if match2_str:
+                        replace_span = (match2_str.span()[0] + match1_str.span()[0], match2_str.span()[1] + match1_str.span()[0])
+                        begining_str = security_php[: replace_span[0]]
+                        after_str = security_php[replace_span[1] :]
+                        security_php = "".join([begining_str, "false", after_str])
+            if security_php:
```
larsks, line 411: This is always going to be `true` (because you set `security_php = file.read()` on line 399, and that isn't guarded by any sort of conditional).

larsks (commented):
I would normally make this sort of change at image build time, rather than at run time. I wouldn't expect the container at runtime to necessarily have write access to these files (from a security perspective, the user under which a web server is running shouldn't be able to modify the files that implement the service). If you're confident it works without additional permission changes, I'm not going to insist on a change.

joachimweyl:
@rob-baron are you clear on what needs to be changed for the two Lars conversations/comments?

xdmod_init.py
rob-baron: ok, I'll take out the conditional.

rob-baron: ok, I'll take out the conditional.

rob-baron: I understand your concerns.  It made more sense to do it in the xdmod-init as I am modifying other parts of the configuration.

rob-baron:
> @rob-baron are you clear on what needs to be changed for the two Lars conversations/comments?

Yes, I do understand what needs to  be changed, just needed to focus on the xdmod-openstack-reporting script this morning as that is where my headspace was.

I am all about getting this to meet other the design sensibilities of the reviewers.

xdmod_init.py
larsks: I think this is a response to [an earlier comment](https://github.com/CCI-MOC/xdmod-cntr/pull/39#pullrequestreview-1010503285), rather than the comment here about the `if` statement.

rob-baron:
I would normally make this sort of change at image build time, rather than at run time. I wouldn't expect the container at runtime to necessarily have write access to these files (from a security perspective, the user under which a web server is running shouldn't be able to modify the files that implement the service). If you're confident it works without additional permission changes, I'm not going to insist on a change.
@larsks I understand your concerns. It made more sense to do it in the xdmod-init as I am modifying other parts of the configuration.

Unfortunately, I had to quote your initial response as kim had placed his comment after your comment and it disrupts the continuity (there is no way to respond to your comment in a thread, unlike the comments that you made when attaching to code.

rob-baron:
I have waited several days for larsks to review, as he must be busy and I have either removed what he objected to and have moved the updating of the php code into to the container build, I am just going to squash and merge.

PR #42, opened 2022-06-27
Mounting empty directories (ephemeral storage) as opposed to relaxing…

rob-baron:
… the permissions in the docker file.

Has the advantage of not having to change the permissions of 3 directories in the docker file, with the disadvantage being that it will not run as a standalone container.


larsks (approved):
This is a good start. There's still a lot going on here that I'm not sure I understand:

- Why are we setting mode 777 on individual executables? Do we expect to write to these at runtime? If so, why?
- Why are we setting mode 777 on `/bin`? Are we installing executables at runtime? 
- Hopefully the changes to `/root` can be dropped when we move to using a different directory for storage.

Anyway, all that can be addressed in a future PR.

PR #47, opened 2022-07-07
Data job

rob-baron:
This is to upload and download config files to the database as a work a round to not having RWX PVs.

### DIFF
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -73,6 +73,7 @@ RUN rm -f /etc/httpd/conf.d/ssl.conf \
 
 COPY ./hypervisor_facts.py /usr/bin/xdmod-openstack-hypervisor
 COPY ./moc_openstack_api_reporting.py /usr/bin/xdmod-openstack-reporting
+COPY ./GetConfigFiles.py /usr/bin/xdmod-get-config-files
 COPY ./xdmod_init.py /usr/bin/xdmod-init 
 RUN chmod 777 /usr/bin/xdmod-openstack-hypervisor \
     && sed -i -e 's/\r$//' /usr/bin/xdmod-openstack-hypervisor \
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -95,7 +95,6 @@ RUN chmod 774 /usr/bin/xdmod-openstack-hypervisor \
     && chgrp -R 0 /etc/xdmod \
     && find /etc/xdmod -type d -exec chmod 764 {} + \
     && find /etc/xdmod -type f -exec chmod 664 {} + \
-    && chmod -R 664 /etc/httpd \
     && chgrp -R 0 /etc/httpd \
     && find /etc/httpd -type d -exec chmod 764 {} + \
     && find /etc/httpd -type f -exec chmod 664 {} + 
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -86,12 +86,15 @@ RUN chmod 777 /usr/bin/xdmod-openstack-hypervisor \
     && chgrp -R 0 /var/lib \
     && chmod -R g+rwX /root \
     && chgrp -R 0 /root \
-    && chmod -R g+rwX /etc/xdmod \
     && chgrp -R 0 /etc/xdmod \
-    && chmod -R g+rwX /etc/httpd \
-    && chgrp -R 0 /etc/httpd 
+    && find /etc/xdmod -type d -exec chmod 764 {} + \
+    && find /etc/xdmod -type f -exec chmod 664 {} + \
+    && chmod -R 664 /etc/httpd \
+    && chgrp -R 0 /etc/httpd \
+    && find /etc/httpd -type d -exec chmod 764 {} + \
+    && find /etc/httpd -type f -exec chmod 664 {} + \
 
-COPY ./assets/allow_insecure_tokens.py /root
+    COPY ./assets/allow_insecure_tokens.py /root
 RUN python3 /root/allow_insecure_tokens.py \
     && rm /root/allow_insecure_tokens.py
 
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -74,6 +74,7 @@ RUN rm -f /etc/httpd/conf.d/ssl.conf \
 COPY ./hypervisor_facts.py /usr/bin/xdmod-openstack-hypervisor
 COPY ./moc_openstack_api_reporting.py /usr/bin/xdmod-openstack-reporting
 COPY ./GetConfigFiles.py /usr/bin/xdmod-get-config-files
+COPY ./xdmod-run-ingestor.sh /usr/bin/xdmod-run-ingestor.sh
 COPY ./xdmod_init.py /usr/bin/xdmod-init 
 RUN chmod 777 /usr/bin/xdmod-openstack-hypervisor \
     && sed -i -e 's/\r$//' /usr/bin/xdmod-openstack-hypervisor \
@@ -81,6 +82,8 @@ RUN chmod 777 /usr/bin/xdmod-openstack-hypervisor \
     && sed -i -e 's/\r$//' /usr/bin/xdmod-openstack-reporting \ 
     && chmod 777 /usr/bin/xdmod-init \
     && sed -i -e 's/\r$//' /usr/bin/xdmod-init \ 
+    && chmod 777 /usr/bin/xdmod-run-ingestor.sh \
+    && sed -i -e 's/\r$//' /usr/bin/xdmod-run-ingestor.sh \ 
     && chmod -R g+rwX /usr/bin \
     && chgrp -R 0 /usr/bin \
     && chmod -R g+rwX /var/lib \
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -76,13 +76,15 @@ COPY ./moc_openstack_api_reporting.py /usr/bin/xdmod-openstack-reporting
 COPY ./GetConfigFiles.py /usr/bin/xdmod-get-config-files
 COPY ./xdmod-run-ingestor.sh /usr/bin/xdmod-run-ingestor.sh
 COPY ./xdmod_init.py /usr/bin/xdmod-init 
-RUN chmod 777 /usr/bin/xdmod-openstack-hypervisor \
+RUN chmod 774 /usr/bin/xdmod-openstack-hypervisor \
     && sed -i -e 's/\r$//' /usr/bin/xdmod-openstack-hypervisor \
-    && chmod 777 /usr/bin/xdmod-openstack-reporting \
+    && chmod 774 /usr/bin/xdmod-openstack-reporting \
     && sed -i -e 's/\r$//' /usr/bin/xdmod-openstack-reporting \ 
-    && chmod 777 /usr/bin/xdmod-init \
+    && chmod 774 /usr/bin/xdmod-init \
     && sed -i -e 's/\r$//' /usr/bin/xdmod-init \ 
-    && chmod 777 /usr/bin/xdmod-run-ingestor.sh \
+    && chmod 774 /usr/bin/xdmod-get-config-files \
+    && sed -i -e 's/\r$//' /usr/bin/xdmod-get-config-files \ 
+    && chmod 774 /usr/bin/xdmod-run-ingestor.sh \
     && sed -i -e 's/\r$//' /usr/bin/xdmod-run-ingestor.sh \ 
     && chmod -R g+rwX /usr/bin \
     && chgrp -R 0 /usr/bin \
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -98,9 +98,10 @@ RUN chmod 774 /usr/bin/xdmod-openstack-hypervisor \
     && chmod -R 664 /etc/httpd \
     && chgrp -R 0 /etc/httpd \
     && find /etc/httpd -type d -exec chmod 764 {} + \
-    && find /etc/httpd -type f -exec chmod 664 {} + \
+    && find /etc/httpd -type f -exec chmod 664 {} + 
 
-    COPY ./assets/allow_insecure_tokens.py /root
+COPY ./assets/allow_insecure_tokens.py /root
+
 RUN python3 /root/allow_insecure_tokens.py \
     && rm /root/allow_insecure_tokens.py
 
--- a/GetConfigFiles.py
+++ b/GetConfigFiles.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 import os
 import sys
+import time
 import argparse
 import json
 import pprint
@@ -105,7 +106,10 @@ def main():
     # config = do_parse_args(config)
 
     write_file_from_db(config, "openstack-cloud-config")
+    time.sleep(20)
     write_file_from_db(config, "xdmod-config")
+    time.sleep(20)
+    write_file_from_db(config, "xdmod-linker")
 
 
 main()
--- a/GetConfigFiles.py
+++ b/GetConfigFiles.py
@@ -1,0 +1,111 @@
+#!/usr/bin/python3
+import os
+import sys
+import argparse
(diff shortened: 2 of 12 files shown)


GetConfigFiles.py
```diff
@@ -0,0 +1,111 @@
...
+import mysql.connector
+
+
+def exec_sql(cursor, sql_stmt, params, error_msg):
+    """executes a simple sql statement"""
+    try:
+        cursor.execute(sql_stmt, params)
+    except Exception as err:
+        print(f"{str(err)} \n {error_msg}")
+        sys.exit()
```
naved001, line 16: is this error handling useful? You are printing the error and then exiting, which is what would happen if you didn't do this at all.

GetConfigFiles.py
```diff
@@ -0,0 +1,111 @@
...
+import pprint
+import mysql.connector
+
+
+def exec_sql(cursor, sql_stmt, params, error_msg):
+    """executes a simple sql statement"""
+    try:
+        cursor.execute(sql_stmt, params)
+    except Exception as err:
+        print(f"{str(err)} \n {error_msg}")
```
naved001, line 15: What is the purpose of passing in an error message `error_msg` to these functions executing sql statements? And then we print the error that we passed? Sorry, not sure if I follow.

GetConfigFiles.py
```diff
@@ -0,0 +1,111 @@
...
+    if data and isinstance(data, list):
+        rec = data[0]
+        with open(rec[0], "wb+") as fp:
+            fp.write(rec[1])
+
+    cnx.commit()
+    cnx.close()
+
+
+def do_parse_args(config):
```
larsks, line 75: This method appears to be unused.

GetConfigFiles.py
```diff
@@ -0,0 +1,111 @@
...
+def main():
+    """the main function"""
+    config = {}
+    if os.path.isfile("./xdmod_init.json"):
+        try:
+            with open("./xdmod_init.json", "r", encoding="utf-8") as file:
+                config = json.load(file)["database"]
+        except IOError:
+            work_dir = os.getcwd()
+            print(f"Ensure the xdmod-init.json file is in the working directory {work_dir}")
```
larsks, line 103: The `except IOError` block will never trigger because of a missing file (because of the `if os.path.isfile(...)` that surrounds the `try/except` block). Probably better just to simplify it to something like:

```suggestion
    try:
        with open("./xdmod_init.json", "r", encoding="utf-8") as file:
            config = json.load(file)["database"]
    except IOError as err:
        work_dir = os.getcwd()
        print(f"Unable to open xdmod_init.json: {err}\nEnsure the xdmod-init.json file is in the working directory {work_dir}")
```

GetConfigFiles.py
```diff
@@ -0,0 +1,111 @@
...
+    # only do this if it is necessary
+    # config = do_parse_args(config)
+
+    write_file_from_db(config, "openstack-cloud-config")
+    write_file_from_db(config, "xdmod-config")
+
+
+main()
```
larsks, line 111: It's better to always use the `if __name__ == "__main__"` guard, even for things that you expect run from the command line. This makes it easier to write tests (because you can import the file without actually executing it).

```suggestion
if __name__ == '__main__':
    main()
```

Dockerfile.moc-xdmod
```diff
@@ -86,12 +87,15 @@ RUN chmod 777 /usr/bin/xdmod-openstack-hypervisor \
...
-    && chgrp -R 0 /etc/httpd 
+    && find /etc/xdmod -type d -exec chmod 764 {} + \
+    && find /etc/xdmod -type f -exec chmod 664 {} + \
+    && chmod -R 664 /etc/httpd \
+    && chgrp -R 0 /etc/httpd \
+    && find /etc/httpd -type d -exec chmod 764 {} + \
+    && find /etc/httpd -type f -exec chmod 664 {} + \
 
-COPY ./assets/allow_insecure_tokens.py /root
+    COPY ./assets/allow_insecure_tokens.py /root
```
larsks, line 98: I don't think you want this indent here (it makes it look like part of the previous `RUN` command, which it is not).

GetConfigFiles.py
larsks: @naved001 these helper functions are all an attempt to avoid a boilerplate `try/except` block around database calls. So e.g., you can `exec_sql(cursor, "SELECT name FROM users WHERE id=?", uid, "Failed to find user")` and then later you can `exec_sql(cursor, "SELECT color FROM widgets WHERE widgetid=?", widgetid, "Failed to find widget")`, and in both cases get a contextually appropriate error message printed to the console.

...but I'm not sure these helper functions are actually necessary or useful; see my comment later on.

GetConfigFiles.py
```diff
@@ -0,0 +1,111 @@
...
+    cursor = cnx.cursor()
+
+    print(f"select file_name, file_data from file_share_db.file where script='{script}'")
+    data = exec_fetchall(
+        cursor,
+        "select file_name, file_data from file_share_db.file where script=%s",
+        (script,),
+        "Unable to select file from db",
+    )
+    pprint.pprint(data)
```
larsks, line 65: We should make debugging statements conditional on some sort of debug flag: we don't want to see this output all the time (because cluttered logs make it harder to spot real problems). Consider using the `logging` module along with appropriate log levels.

larsks (changes requested):

joachimweyl:
@rob-baron what are the next steps for merging this PR?

GetConfigFiles.py
rob-baron: asdf

rob-baron:
@joachimweyl - apparently I started a review 16 days ago saying that I would be making the changes, so all of my comments were in draft and only viewable by me (quite annoying).  As I have told you multiple times, I have made the changes, I just haven't pushed them as I haven't finished testing them yet.

GetConfigFiles.py
rob-baron: I have removed many of the long ones.  I do however like to see checkpoints in the log file in order that I can follow the code as it is working, not just logging errors.  This way I have at least a decent chance of viewing a log file and understanding if it is behaving normally.

rob-baron:
@joachimweyl, some of the commits were adding the kubernetes job definition, I have committed the code that addresses the comments because I just want this to be done with.  I had been putting this off as I have been focused on getting the data in - and during work hours I have been working at figuring out what data STF provides in terms of volumes events.  But since this PR came up multiple times today - I decided to address it during my free time tonight.

@larsks - I have made most of the changes that offended your design sensibilities.

GetConfigFiles.py
naved001: > Why do I want to dig though the call stack when the sql statment has a simple error?

Sure, I see your point. But I guess personally I am not so bothered by a stack trace I tend to look at the last error first before going up the trace.

GetConfigFiles.py
```diff
@@ -0,0 +1,72 @@
...
+        with open("/etc/xdmod/xdmod_init.json", "r", encoding="utf-8") as file:
+            config = json.load(file)["database"]
+    except IOError:
+        print("Ensure the xdmod-init.json file is in /etc/xdmod/xdmod_init.json")
+
+    cnx = connect_to_db(config)
+    cursor = cnx.cursor()
+
+    write_file_from_db(cursor, "etc-xdmod")
+    with subprocess.Popen(["/usr/bin/base64", "-d", "/etc/xdmod/etc_xdmod.b64"], stdout=subprocess.PIPE, stderr=subprocess.PIPE) as base64:
```
naved001, line 60: I **think** we should avoid using subprocesses if its not strictly required. The base64 python library is fairly simple to use.
```
>>> import base64
>>> encoded = base64.b64encode(b"hello world")
>>> print(base64.b64decode(encoded))
b'hello world'
```

GetConfigFiles.py
```diff
@@ -0,0 +1,72 @@
...
+        b64_out, b64_err = base64.communicate()
+        print(f"errors: {b64_err}")
+        with open("/etc/xdmod/etc_xdmod.tgz", "wb") as fptr:
+            fptr.write(b64_out)
+        os.system("/usr/bin/tar -xzf /etc/xdmod/etc_xdmod.tgz")
```
naved001, line 65: probably be a better idea to use https://docs.python.org/3/library/tarfile.html

xdmod_init.py
```diff
@@ -396,5 +403,16 @@ def main():
         os.popen("/usr/share/xdmod/tools/etl/etl_overseer.php -p ingest-resource-types")
         os.popen("/usr/share/xdmod/tools/etl/etl_overseer.php -p ingest-resources")
 
+        cnx = connect_to_db(xdmod_init_json["database"])
+
+        os.system("tar -czf /etc/xdmod/etc_xdmod.tgz /etc/xdmod/*")
+        os.system("base64 /etc/xdmod/etc_xdmod.tgz > /etc/xdmod/etc_xdmod.b64")
+        write_file_to_db(cnx.cursor(), "/etc/xdmod/etc_xdmod.b64", "etc-xdmod")
+        os.system("rm /etc/xdmod/etc_xdmod.tgz")
+        os.system("rm /etc/xdmod/etc_xdmod.b64")
```
naved001, line 412: os.remove("/etc/xdmod/etc_xdmod.64")
https://docs.python.org/3/library/os.html#os.remove

xdmod_init.py
```diff
@@ -1,12 +1,9 @@
 #!/usr/bin/python3
-# pylint: disable=line-too-long broad-except too-many-locals too-many-branches
+# pylint: disable=line-too-long broad-except too-many-locals too-many-branches too-many-statements
```
naved001, line 2: You can probably remove broad-except

joachimweyl:
@rob-baron it is very exciting that this is so close to done. I only see a few change requests left. Nice work.

larsks (changes requested):
Most of the review comments so far have been about the *code*, but I'd like to review the organization of the pull request. Unfortunately, GitHub doesn't provide great tools for that, so I've had to get a little creative.

First, some general comments:

1. The commits in this PR don't follow the conventions that we've called out [here](https://github.com/CCI-MOC/ops-docs/blob/master/policies/0007-pull-requests.md):

    - There are several instances where commit messages fail to describe the reason for their changes (see below for why this matters).

    - There are commits where the message is one that we've explicitly asked people to avoid. From the same document I linked above:

        > Commit messages that should explicitly be avoided are those that provide no information about the contents of the commit, such as:
        > 
        > - "Responding to review feedback"
        > - "Integrated changes from Bob"
        > - "Refactored code"

      Additionally, there was at least "fixed a typo" commit that should simply have been folded into whichever commit originally introduced the change.

1. There are several commits that contain unrelated changes that should be split into multiple commits.

1. The relationship between the purported purpose of this pull request ("This is to upload and download config files to the database as a workaround to not having RWX PVs") and some of the commits is unclear: it seems like they should perhaps have been part of a different pull request (specifically, the changes to Kubernetes manifests to introduce development overlays, the ingestor CronJob, etc, don't really seem to fit the PR description).

---

There are several advantages to keeping individual commits tightly focused. From a code maintenance perspective, it makes it easier for someone to answer the question, "Why did we make this change?". Looking at a block of code, we can use `git blame` to find the commit responsible for the change; if the commit was focused and the commit message is relevant, we're done: we'll have a good idea what the rational was for the code at which we're looking.

If, on the other hand, the commit message says something like "Responding to review comments", we have a lot more work: first, we need to find out from which pull request the commit originated, and then we have to page through the pull request trying to figure *which* review comments the changes is addressing.

Similarly, keeping a *pull request* focused on a particular topic helps both the developer -- because it allows them to focus their efforts on a particular topic -- but also the reviewer, because it doesn't require them to context-switch between unrelated parts of the codebase. Because a focused pull request is easier to review, it's also going to get merged quicker, allowing the developer to continue work that is built upon the changes in the PR.

Of course, our brains don't necessarily operate in a linear fashion, which means our development doesn't always follow a path that maps directly into a set of clean pull requests. It's important to realize that *the commits we present in a pull request don't need to be the same as the commits we make while developing the code*. We can follow our own path during development; as long as we keep our individual commits relatively small and contained, it's relatively easy to re-order them, squash them, and otherwise massage them into something more reviewable at the time we open a pull request.

---

Rather than simply leaving a bunch of comments, I spent some time this past week refactoring the pull request (not the code, just the PR structure) to try to model what I'm describing here. I've split the changes into three separate pull requests:

1. PR #51  has the changes related to using the database for fetching and storing configuration files (`GetConfigFiles.py`, etc).
1. PR #52  has adds the ingestor `CronJob`
1. PR #53  adds the development overlays and `BuildConfigs`.

When all three pull requests are merged, they contain the *exact* same code as in this pull request. There are some minor whitespace differences due to some commits being merged and/or removed:

```
diff --git a/Dockerfile.moc-xdmod b/Dockerfile.moc-xdmod
index 24109e9..8db618b 100644
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -95,10 +95,9 @@ RUN chmod 774 /usr/bin/xdmod-openstack-hypervisor \
     && find /etc/xdmod -type f -exec chmod 664 {} + \
     && chgrp -R 0 /etc/httpd \
     && find /etc/httpd -type d -exec chmod 764 {} + \
-    && find /etc/httpd -type f -exec chmod 664 {} + 
+    && find /etc/httpd -type f -exec chmod 664 {} +
 
 COPY ./assets/allow_insecure_tokens.py /root
-
 RUN python3 /root/allow_insecure_tokens.py \
     && rm /root/allow_insecure_tokens.py
```

When looking at these pull requests, note that:

1. Each pull request contains an independent set of changes. Each PR can be reviewed and merged without conflicts with the other PRs.
1. Large commits with non-descriptive messages have been split into multiple smaller commits with descriptive messages.
1. Typo fixes have been rolled into existing commits.
1. There are no merge commits in the pull requests and they will all apply cleanly against the current `main` branch.
1. All the commit messages have been updated to follow common conventions.

There are a few commits that are marked `[FIXME]` in those #51 ; these are situations in which I wasn't able to intuit the reason for a change, so they require updated verbiage for both the subject and the body of the message before the changes are merged.

joachimweyl:
@larsks wow, thank you for the time you put into this. I just learned a lot about how I can enhance my GitHub pulls. Thank you for this learning opportunity.

rob-baron:
@larsks 

I decided against splitting this up as:
1) Things were taking longer to get through the review process
2) Mainn needed to be unblocked to move forward
3) The other parts required changing a similar set of files of which if the other parts also required code changes, that would mean extract work to resolve merge conflicts.

And you are still requesting changes - despite adding in all of the changes you requested - or are you requesting more changes?

larsks:
> Things were taking longer to get through the review process

This exactly why it's a good idea to split things up: there's a much better chance that the pull requests can get reviewed and merged in a timely fashion when they are smaller and more focused.

> And you are still requesting changes - despite adding in all of the changes you requested - or are you requesting more changes?

I am requesting changes in how this pull request is structured: specifically, in the content and the organization of the commits.

We can simply merge the three new pull requests (#51, #52, and #53) once you've addressed the `[FIXME]` commit messages and the comments I've left on the pull requests, or you can replace the contents of this PR with the merged content of those three pull requests if you'd prefer a single PR.

I'd be be happy to help with either of those steps. Nothing here involves a great deal of work or time, so we should be able to finish up today.

larsks:
It may be useful to check out pull requests locally -- both to more fully understand the changes I've made, and also to make your own changes to the commit messages, etc. The easiest way to do that is using the [github cli](https://cli.github.com/); from within your working directory for this repository you can run, e.g:

```
gh pr checkout 51
```

rob-baron:
@larsks, I already checkout pull requests locally as it is the only rational manner to deal with them.  I am more concern with resolving merge conflicts which, given the files involved will probably be messy (not necessarily the yaml file though as those are pretty contained).

larsks:
There shouldn't be any particular merge conflicts here: #52 and #53 are completely independent, and #51 should apply correctly if #47 did.

rob-baron:
@larsks, ok, I was figuring that this needed to be merged with the changes suggested by naved - but I will do as you suggest and close #47 and merge 51 and be done with that.

rob-baron:
This is closed as suggested by @larsks

PR #48, opened 2022-07-15
[WIP] Add openshift-xdmod job

tzumainn:


Dockerfile.moc-xdmod
```diff
@@ -89,6 +90,14 @@ RUN chmod g+rwX /usr/bin/xdmod-openstack-hypervisor \
     && chmod -R g+rwX /etc/httpd \
     && chgrp -R 0 /etc/httpd 
 
+# openshift reporting
+# checkout https://github.com/OCP-on-NERC/xdmod-openshift-scripts
+# copy openshift xdmod script to /usr/bin/xdmod-openshift.sh (to be created)
```
rob-baron, line 95: I would go with /usr/bin/xdmod-openshift-reporting as that seems to fit with their naming convention (just considering all of the xdmod-... scripts that exist in /usr/bin.

Dockerfile.moc-xdmod
```diff
@@ -89,6 +90,14 @@ RUN chmod g+rwX /usr/bin/xdmod-openstack-hypervisor \
...
+#   # run oc login
+#   # set OPENSHIFT_PROMETHEUS_URL
+#   # python openshift_metrics/openshift_prometheus_metrics.py --output-file <output>
+#   # /usr/bin/xdmod-shredder -f slurm -i <output> -r <resource>
+
```
rob-baron, line 100: now I understand you notation. So:
1) I would set the URL in the job
2) you will need to down load the configuration files from the database as the volume storing them cannot be shared (it is RWO and not RWX as NESSE(?) does not support RWX).

joachimweyl:
@tzumainn were you able to get the information you needed. How close are we to a merge of this PR?

joachimweyl:
@tzumainn can you reply to the thread above with the resolution/example for step 2?
@rob-baron what part of your method is not fully working?

tzumainn:
> @tzumainn can you reply to the thread above with the resolution/example for step 2? @rob-baron what part of your method is not fully working?

Let me back up: the hold up right now is that Rob is (correctly, and supported by the advice of others) containerizing XDMoD and then creating kubernetes jobs to run the required XDMoD periodic tasks. Neither of us are experts here, I don't think, so doing all of this is not quick.

The alternative I mentioned would be to simply stand up XDMoD uncontainerized, and use something like cron jobs to run the scripts. That should be relatively easy.

If things become extremely urgent, I think the alternative would be fairly quick (and I imagine the data could be exported into the containerized solution once that was ready). However I also think we may be close to solving the containerized deployment, which would be far preferable.

rob-baron:
@tzumainn Yes, standing up xdmod using cron jobs and the database in the same system on bare metal is the back up plan.  However, I feel that we are close to getting this up and running under openshift.

@joachimweyl To answer your question, right now, there are 2 current issues:
1) Why is the database dropping connections? 
      a) sometimes causes the xdmod-setup/xdmod-init scripts to fail
      b) seems to cause the interface to stop working
      c) causes the kubernetes cron job to fail  
2) Why is the kubernetes cron job starting up more than 1 job when the job fails? 

So far today, I have been looking into STF, attempting to find the API and log into it.  I am wondering if it depends on ceilometer to collect the data, or if it is collecting the data similar to ceilometer (for the volume data and potentially the compute data - as this will, potentially, be easier).

My impression as to my priorities is to:
1) get the data collected so that it can be displayed in xdmod
2) work on the example to reproduce the database dropping the connection
3) any other issues (including the aforementioned kubernetes cron job)

Dockerfile.moc-xdmod
rob-baron: Here (https://github.com/CCI-MOC/xdmod-cntr/pull/47/files#diff-507b1af3802214396997dcfc436e9c9e62f853102bea79b1fe6e4c030d34c735).  I thought I pointed this out to you a week or so ago.

Hopefully, we get RWX volumes on OpenShift and the kludge will go away!

tzumainn: Yep, you definitely did! Sorry, this comment was from a while ago.

You may see a lot of updates to this PR, only because I'm pushing updates to my local branch while testing various things out (I just managed to get the updated build config working). Feel free to ignore everything - I'll let you know when I think this is ready for another look!

tzumainn:
@joachimweyl not all that close, I don't think! I'll emphasize: everything I'm doing is based on what rob is doing, so our fates are a bit bound together.

I'm actually going to close this PR because I'm experimenting with a few things that require me to constantly push changes onto my branch, but I don't want to create too much noise here!

PR #51, opened 2022-08-01
Use database for storing/fetching config files in lieu of RWX storage

larsks:
This is to upload and download config files to the database as a workaround for not having RWX PVs.

### DIFF
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -84,10 +84,12 @@ RUN chmod g+rwX /usr/bin/xdmod-openstack-hypervisor \
     && chgrp -R 0 /usr/bin \
     && chmod -R g+rwX /root \
     && chgrp -R 0 /root \
-    && chmod -R g+rwX /etc/xdmod \
     && chgrp -R 0 /etc/xdmod \
-    && chmod -R g+rwX /etc/httpd \
-    && chgrp -R 0 /etc/httpd 
+    && find /etc/xdmod -type d -exec chmod 764 {} + \
+    && find /etc/xdmod -type f -exec chmod 664 {} + \
+    && chgrp -R 0 /etc/httpd \
+    && find /etc/httpd -type d -exec chmod 764 {} + \
+    && find /etc/httpd -type f -exec chmod 664 {} +
 
 COPY ./assets/allow_insecure_tokens.py /root
 RUN python3 /root/allow_insecure_tokens.py \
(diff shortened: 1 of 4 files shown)


xdmod-run-ingestor.sh
```diff
@@ -2,4 +2,5 @@
 
 cp /mnt/xdmod_init/xdmod_init.json /etc/xdmod/xdmod_init.json
 /usr/bin/xdmod-get-config-files 
+/usr/bin/sleep 30
```
larsks, line 5: @rob-baron Was this related to the database connection dropping issue?

Dockerfile.moc-xdmod
```diff
@@ -76,14 +76,15 @@ COPY ./moc_openstack_api_reporting.py /usr/bin/xdmod-openstack-reporting
 COPY ./GetConfigFiles.py /usr/bin/xdmod-get-config-files
 COPY ./xdmod-run-ingestor.sh /usr/bin/xdmod-run-ingestor.sh
 COPY ./xdmod_init.py /usr/bin/xdmod-init 
-RUN chmod g+rwX /usr/bin/xdmod-openstack-hypervisor \
+RUN chmod 774 /usr/bin/xdmod-openstack-hypervisor \
```
larsks, line 79: @rob-baron in this change, you're replacing symbolic modes like `g+rwX` with numeric modes; what problem was this solving?

GetConfigFiles.py
```diff
@@ -39,8 +39,8 @@ def write_file_from_db(cursor, script):
         if not os.path.isdir(directory):
             print(f"Creating Direcotry {directory}")
             os.makedirs(directory)
-        with open(file, "wb+") as fp:
-            fp.write(rec[1])
+        with open(file, "wb+") as file:
+            file.write(rec[1])
```
larsks, line 43: I'm not sure this change makes sense: with this change, you're shadowing the `file` variable, so what was originally a string is now a an open file object. It will *work*, because Python is a dynamically typed language, but I think the original version was more clear.

xdmod-run-ingestor.sh
rob-baron: @larsks This was thrown in as an attempt to work around the database connection dropping issue.  

2 things:

1) I am not certain if this works
2) This cannot be here long term as it will not meet the overall performance requirements

I have kept it here as I am developing this as it was useful for development.

GetConfigFiles.py
rob-baron: ok will change.

larsks: The easiest solution here is to simply delete the commit that contains this change. I've done that and updated this PR.

larsks:
I think this can be merged!

PR #52, opened 2022-08-01
Add CronJob for running xdmod ingestor

larsks:
Run the xdmod ingestor daily in order to import new data.

### DIFF
--- a/k8s/kube-base/job-xdmod-ingestor.yaml
+++ b/k8s/kube-base/job-xdmod-ingestor.yaml
@@ -1,0 +1,40 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: xdmod-ingestor
+spec:
+  schedule: "@daily"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: moc-xdmod
+              image: moc-xdmod
+              imagePullPolicy: IfNotPresent
+              command:
+                - "/usr/bin/xdmod-run-ingestor.sh"
+              volumeMounts:
+                - name: vol-xdmod-conf
+                  mountPath: /etc/xdmod
+                - name: vol-clouds-yaml
+                  mountPath: /etc/openstack
+                - name: vol-xdmod-init
+                  mountPath: /mnt/xdmod_init
+                - name: vol-var-log-xdmod
+                  mountPath: /var/log/xdmod
+          restartPolicy: Never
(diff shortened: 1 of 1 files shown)


k8s/kube-base/job-test.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: batch/v1
```
larsks, line 1: @rob-baron  I don't think you need this `Job`; you can [spawn a Job from a CronJob manually](https://www.craftypenguins.net/how-to-trigger-a-kubernetes-cronjob-manually/) using e.g.:

```
kubectl create job --from=cronjob/xdmod-ingestor test-xdmod-ingestor
```

k8s/kube-base/job-xdmod-ingestor.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: xdmod-ingestor
+spec:
+  schedule: "* * */1 * *"
```
larsks, line 6: This schedule is going to run the job every minute of every day (because you have `*` for minute (and hour, and month, and day-of-week, and `*/1` means "every one day"). If your intent is to run this once/day, you can write (from [the docs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax)):

```
schedule: "@daily"
```

Or if you want a time other than midnight, specify the minutes and hour explicitly. The following will run at 1:00AM every day:

```
schedule: "0 1 * * *"
```

rob-baron: When I wrote that, I wanted every 5 minutes, but hadn't settled on that.  however, for testing purposes, every minute was good, though It needs to be once a day or '0 0 * * *'

k8s/kube-base/job-test.yaml
rob-baron: I included this for Mainn, as it is a way to just test the job.  will delete.

larsks: I've removed that commit from the PR.

k8s/kube-base/job-xdmod-ingestor.yaml
larsks: I've updated the cronjob to use the `@daily` schedule.

larsks:
I think this can be merged!

PR #56, opened 2022-08-03
Add openshift-xdmod-staging job

tzumainn:


k8s/kube-base/deployment-xdmod.yaml
```diff
@@ -104,6 +110,9 @@ spec:
         - name: vol-xdmod-data
           persistentVolumeClaim:
             claimName: pvc-xdmod-data
+        - name: vol-xdmod-openshift-data
+          persistentVolumeClaim:
+            claimName: pvc-xdmod-openshift-data
```
rob-baron, line 115: I don't think we can have this here as we don't have RWX PVCs yet.  Your job will work if the job is started on the same node, but should fail if it is started on other nodes.  This might be why it was started 3 times.

k8s/kube-base/job-xdmod-openshift-staging.yaml
```diff
@@ -0,0 +1,58 @@
...
+                - name: vol-xdmod-conf
+                  mountPath: /etc/xdmod
+                - name: vol-xdmod-src
+                  mountPath: /usr/share/xdmod
+                - name: vol-xdmod-init
+                  mountPath: /mnt/xdmod_init
+                - name: vol-var-log-xdmod
+                  mountPath: /var/log/xdmod
+                - name: vol-xdmod-openshift-data
+                  mountPath: /root/xdmod_openshift_data
```
rob-baron, line 38: Let's use a different directory other than root for this.  Let's create a /data directory as previous review comments view the /root directory as special.

rob-baron (changes requested):
It looks good.  lets try moving the data directory from /root/... to /data and not referencing the PVC for the openshift data in the UI deployment.

k8s/kube-base/deployment-xdmod.yaml
tzumainn: Ah, okay! I've removed this. The three times thing was actually something different though; my initial cron job scheduler was incorrect, and configured to run every minute!

k8s/kube-base/job-xdmod-openshift-staging.yaml
tzumainn: Sounds good - done!

tzumainn:
I've updated the PR - let me know if there are other issues!

k8s/kube-base/deployment-xdmod.yaml
rob-baron: That also makes sense.

rob-baron (approved):
This looks good.

Dockerfile.moc-xdmod
```diff
@@ -97,6 +98,22 @@ RUN chmod 774 /usr/bin/xdmod-openstack-hypervisor \
...
+
+# openshift reporting
+COPY ./xdmod-openshift-reporting.sh /usr/bin/xdmod-openshift-reporting.sh
+RUN git clone https://github.com/OCP-on-NERC/xdmod-openshift-scripts \
+    && cd xdmod-openshift-scripts \
+    && pip install -r requirements.txt \
+    && chmod g+rwx /usr/bin/xdmod-openshift-reporting.sh
+
+# force downgrade of mysql-connector-python
+RUN pip3 install -I mysql-connector-python==8.0.28
```
knikolla, line 115: Just for reference, what is the reason for this?

Dockerfile.moc-xdmod
```diff
@@ -97,6 +98,22 @@ RUN chmod 774 /usr/bin/xdmod-openstack-hypervisor \
     && find /etc/httpd -type d -exec chmod 764 {} + \
     && find /etc/httpd -type f -exec chmod 664 {} +
 
+# openshift client
+RUN wget https://downloads-openshift-console.apps.ocp-staging.massopen.cloud/amd64/linux/oc.tar --no-check-certificate \
```
knikolla, line 102: Why are you downloading it from the staging cluster? I'm hesitant to introduce a dependency to a staging cluster in the build process of the container.

Dockerfile.moc-xdmod
tzumainn: Ah, there's a bug in 8.0.30 that makes mariadb unable to recognize the utf8 charset.

Dockerfile.moc-xdmod
tzumainn: Fair - is there a better place to download it from? The openshift documentation didn't have an ideal public repository, but I could be missing something!

knikolla: I'm getting it from here https://github.com/CCI-MOC/openshift-acct-mgt/blob/fa396d7752ecf21248071f022ae5e43c1ae24176/.github/workflows/functional-tests.yaml#L35-L36

Dockerfile.moc-xdmod
knikolla: Do you have a link to the bug so that we can track when it's resolved and can remove the workaround?

joachimweyl:
@tzumainn what are next steps for resolving these change requests?

tzumainn:
@joachimweyl oh, I just have to do them :) Yesterday was a catch-up day for a bunch of stuff after PTO; I'll try and get this done today.

Dockerfile.moc-xdmod
tzumainn: Added a link to the issue; there doesn't seem to be a direct bug report that I can find.

Dockerfile.moc-xdmod
tzumainn: Ah, thanks - updated!

tzumainn:
I've addressed the comments. Rob, I took a look at the jobs running while I was on PTO and I noticed that there were occasional database connection issues when shredding; do you know if there's a way to address that?

rob-baron:
>I've addressed the comments. @rob-baron, I took a look at the jobs running while I was on PTO and I noticed that there >were occasional database connection issues when shredding; do you know if there's a way to address that?

That is exactly the problem.  it is intermittent and sometime I get connection issues.  What I have done to address this is to place strategic sleeps to allow either the database to catch up.

tzumainn:
@rob-baron ah, do you have an example I could see? should I put a sleep command into the xdmod-openshift-reporting.sh script?

joachimweyl:
@tzumainn were you and @rob-baron able to connect on the example code?

tzumainn:
@joachimweyl I haven't heard back yet! I figure he's been working on other stuff, but the sleep issue is really the last open problem with the PR, I think!

rob-baron:
@joachimweyl @tzumainn 

So where I have done this with some success is in the xdmod-init process here (https://github.com/CCI-MOC/xdmod-cntr/blob/main/xdmod_init.py) look around 296 - I named the variable "nap_time".  

I don't think that this will work here as you are not hitting the database when you are collecting the data.

What I have also done is to break large files into smaller files.  For example, instead of having 1 large file containing all of the events, I have broken it into daily event files.  You might consider this depending on the size of file that you are trying to shred - that is, is the initial file containing more than a day's worth of data?

joachimweyl:
@tzumainn what are the next steps for you on this PR?

tzumainn:
@joachimweyl I'm testing various fixes for the issue, but none of them seem to work consistently so far. The amount of data being processed is far from large, so I don't think Rob's suggestion applies!

rob-baron:
@tzumainn This is exactly the problem that I have been telling you about.  I have this same problem.  In all honestly, I don't know what the root cause is.  It is my feeling that the database is not operating at the speed that it should be.  I suspect the underlying issue is the interaction between kubernetes and ceph - but I have no way to either prove or disprove this suspicion.

I have used postgress on openshift 3.11 without this issue, I have also used mysqldb under openshift 3.11 without this issue without having to tune the database (ie adjust any timeouts).  I have increased the size of the data volume of the database, so I don't believe that is the issue.  Quite frankly, the data is so small especially since I get the unable to connect when initializing the database(s) using xdmod-setup.

When I have brought this up in our meetings I am told that I need to provide an example that shows this problem.  I have started a project (https://github.com/CCI-MOC/dbex) to do so.  You're welcome to contribute to that endeavor if you have time.

I thought that I have told you this before.

Even though this doesn't work consistently yet as it fails to connect with the mariadb database, I suggest we merge this and consider working on figuring out the root cause as opposed to working around the problem.

You're also welcome to run any database benchmarks.  At this point, we might get lucky and see from benchmarking the performance ways to modify the example project to replicate this issue.

Come to think of it, another thing to try is to swap mariadb with mysql.  Although I have the impression mariadb and mysql are not 100% compatible, this is something that is easy and might work.

tzumainn:
@rob-baron yep! you definitely told me before; I just thought that you had found a final solution, but it sounds like you're in the same place I am! I'm currently setting my cron jobs to sleep for 21 hours after running so I can rsh in and poke around a bit. What I've found is that I can spam just the xdmod-shredder data and it will extremely inconsistently pass or fail. I am currently thinking of updating my script to check the out of xdmod-shredder, and if it hits an error code, to simply try again.

rob-baron:
@tzumainn have you checked your logs.  The logs that I looked at didn't inform me of anything other than the "lost+found" warning.  We should probably compare notes at some point.

rob-baron:
@tzumainn that is a work-a-round that would work for the shredding and the ingestion.  We could restart the script if the "connection" error message appears.  How many tries till it just gives up?

I started with the nerc 0 infra cluster and that exhibits this problem with a higher frequency.

Unfortunately, it will not work for the UI which has the same problem.

tzumainn:
Ah, if the UI has this issue as well, then I don't really see a reasonable workaround other than to not deploy xdmod this way. But maybe someone with more experience than I do would have a better idea!

rob-baron:
@tzumainn why don't you deploy xdmod on a VM.

I will work on the openshift install for 2 reasons:
1) This is really how we should deploy xdmod on the nerc.
2) This may lead to uncovering an issue our users will hit, and I would prefer to deal with problems like this sooner than later.

Shall we squash and merge this PR?  I am willing to as long as it works when you can connect.

tzumainn:
@rob-baron Ah, sorry, I just meant if the UI issue you're working on can't be resolved. I'm pretty sure this iteration of the PR will work; I'm just waiting for the cronjob to pass a time or two with the expected result to be sure! I'll let you know when that happens.

tzumainn:
... and it seems to work! You can look at the output at https://console-openshift-console.apps.ocp-staging.massopen.cloud/k8s/ns/xdmod-tzumainn/pods/xdmod-openshift-staging-job-27686940--1-7snkm/logs; it fails to connect to the database twice before successfully shredding. So yeah, I think this PR can be merged. It should already be rebased from the main branch; let me know if there are any other changes it needs!

PR #64, opened 2022-08-23
Reorganization to support deploying with ArgoCD

rob-baron:
This adds an overlay that ArgoCD can point to in order to deploy onto the NERC.

This overlay may be moved out of this project and into a project that is specific to the NERC as it will contain configuration that is specific to the NERC.

This is meant to work with (https://github.com/OCP-on-NERC/nerc-ocp-apps/pull/8)


k8s/overlays/nerc-ocp-infra/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+bases:
```
naved001, line 1: `bases` has been deprecated for a while: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/bases/

it should be changed to `resources`.

naved001 (commented):
@rob-baron Thanks for arranging the changes commit-by-commit with useful commit messages.

In one of the commits you set `mysql-connector-python` to a specific version, but then after the merge commit you unset it. I don't understand that change. Did something change that you no longer needed to pin `mysql-connector-python` to version 8.0.29?

knikolla:
> @rob-baron Thanks for arranging the changes commit-by-commit with useful commit messages.
> 
> In one of the commits you set `mysql-connector-python` to a specific version, but then after the merge commit you unset it. I don't understand that change. Did something change that you no longer needed to pin `mysql-connector-python` to version 8.0.29?

++ on the breaking things down into commits with explanations for each.

For the pin, it's still present on the file, but much later and with a `-I` tag which forces the downgrade. I'm okay with the way it's written now, with the first `pip install` being unversioned and the second one being versioned.

https://github.com/CCI-MOC/xdmod-cntr/blob/4be0bc3ec2b03902ab5e4d04537510bbf5d06067/Dockerfile.moc-xdmod#L114-L116

k8s/overlays/nerc-ocp-infra/kustomization.yaml
rob-baron: ok changed bases to resources.  Had to update my openshift client (oc) in order to support this change.

rob-baron:
Somewhere, between the first pip3 install of the mysql-connector and the the "pip3 install -I" of the mysql-connector somehow the mysql-connector gets updated.  Not completely sure why, however with the pip3 install -I being placed after everything else is installed works, so I went with it.

joachimweyl:
@naved001 I see your CR was resolved, do you have any other CRs?

naved001 (approved):

k8s/xdmod-build/bc-xdmod-dev.yaml
```diff
@@ -8,7 +8,6 @@ spec:
     type: Git
     git:
       uri: https://github.com/CCI-MOC/xdmod-cntr.git
-      ref: XdmodK8sDeployment
```
larsks, line 11: Should this be `ref: main` instead so that it's clear what branch it's using by default (and similarly for the other buildconfig)?

Dockerfile.moc-xdmod
```diff
@@ -62,7 +62,7 @@ WORKDIR /root
 RUN mkdir -p /root/rpmbuild/RPMS/noarch \
     && mkdir /tmp/supremm \
     && pip3 install --upgrade pip \
-    && pip3 install keystoneauth python-keystoneclient python-novaclient python-cinderclient cryptography python-openstacksdk mysql-connector-python pexpect
+    && pip3 install keystoneauth python-keystoneclient python-novaclient python-cinderclient cryptography python-openstacksdk mysql-connector-python==8.0.29 pexpect
```
larsks: With all these requirements, it would be better to move them into a `requirements.txt` file in the repository.

larsks (changes requested):
I would like to rebase this on `main` to remove the merge request and remove commits ee88922 and 086d17e, which collectively result in no changes. I'm happy to do this if you would prefer.

k8s/xdmod-build/bc-xdmod-dev.yaml
rob-baron: It should be eliminated as it should default to the main branch.

joachimweyl:
@larsks your CR has been outdated by updates. Is this ready to be approved?

k8s/overlays/nerc-ocp-infra/patches/bc-xdmod.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: build.openshift.io/v1
+kind: BuildConfig
+metadata:
+  name: bc-xdmod
+spec:
+  source:
+    type: Git
+    git:
+      ref: ArgoDeploy
```
larsks, line 9: Should this be updated to `main` before we merge?

k8s/xdmod-build/bc-xdmod-dev.yaml
```diff
@@ -8,15 +8,13 @@ spec:
...
     contextDir: /
   strategy:
     type: Docker
     dockerStrategy:
       dockerfilePath: Dockerfile.moc-xdmod.dev
       from:
         kind: ImageStreamTag
-        name: moc-xdmod:latest
```
larsks, line 19: Don't we need an imagestream name here?

larsks:
@joachimweyl I've left two minor questions.

k8s/overlays/nerc-ocp-infra/patches/bc-xdmod.yaml
rob-baron: This should be updated to the specific commit that we will use to deploy from

k8s/xdmod-build/bc-xdmod-dev.yaml
rob-baron: Is this not a image stream?  As in the following:

        >oc get is
        NAME            IMAGE REPOSITORY                                                       TAGS     UPDATED
        moc-xdmod       image-registry.openshift-image-registry.svc:5000/xdmod/moc-xdmod       latest   6 days ago
        moc-xdmod-dev   image-registry.openshift-image-registry.svc:5000/xdmod/moc-xdmod-dev   latest   5 days ago

larsks: Right, but in this commit, you're *removing* the `name:` element.

larsks: Yeah, I was misreading things. Neeeever mind.

larsks (approved):

Issue #63, opened 2022-08-23
Load the mariadb config file into the container


rob-baron:
Create a config map with the mariadb config file and mount the file in the appropriate location in the mariadb container.

This will enable tuning certain parmeters, such as some timeouts that I suspect we are hitting as the pattern seems to be that we are able to log in, then we are not able to log in for 20 seconds, then we are.  This could be for a couple of reasons:

1) not enough concurrent connections
2) timeingout during the log in 

As this happens even in the install, when we are creating the databases, and there is only 1 user, this kinda rules out not enough concurrent connections.  Besides, the last that I checked, the concurrent connections were set to 50.

What I am suspecting more is that something is timing out.  When the connection is requested, the database needs to fetch the user data from cache, if it is not there, it will fetch from the mounted volume.  if the volume is slow, then the connection request times outs but the data is still caches.  5-10 seconds later, when the next connection is requested, the cache has the data and the connection is created.

rob-baron:
1)  To start with running this command in the mariadb pod/container:
           mysqld --help --verbose
    gives the following configuration files:
           /etc/my.cnf /etc/mysql/my.cnf ~/.my.cnf 
    which mariadb reads in order.
    
2)  neither  /etc/my.cnf  nor ~/.my.cnf exist

3) the following configuration files exist
        /etc/mysql/debian-start  
        /etc/mysql/debian.cnf	
        /etc/mysql/mariadb.cnf	
        /etc/mysql/my.cnf
        /etc/mysql/conf.d/mysqld_safe_syslog.cnf
        /etc/mysql/mariadb.conf.d/

4) The three that seem to be important for this are 
        /etc/mysql/mariadb.cnf	
        /etc/mysql/my.cnf
        /etc/mysql/conf.d/mysqld_safe_syslog.cnf

5) as  /etc/mysql/conf.d/mysqld_safe_syslog.cnf only contains:
        [mysqld_safe]
        skip_log_error
        syslog
        
     I will include it for completeness, but it is not that important.

PR #65, opened 2022-08-25
Use configmaped files for mariadb configuration

rob-baron:
This resolves (https://github.com/CCI-MOC/xdmod-cntr/issues/63)

To summarize, using configmaps for the mariadb configuration allows
for tuning parameters to adjust for the enfironment that mariadb is
running in.

### DIFF
--- a/k8s/kube-base/deployment-mariadb.yaml
+++ b/k8s/kube-base/deployment-mariadb.yaml
@@ -25,7 +25,34 @@ spec:
           volumeMounts:
             - name: vol-mariadb
               mountPath: /var/lib/mysql
+            - name: vol-my-conf
+              mountPath: /etc/mysql/my.cnf
+              subPath: my.cnf
+            - name: vol-mariadb-conf
+              mountPath: /etc/mysql/mariadb.cnf
+              subPath: mariadb.cnf
+            - name: vol-mysql-safe-conf
+              mountPath: /etc/mysql/conf.d/mysqld_safe_syslog.cnf
+              subPath: mysqld_safe_syslog.cnf
       volumes:
         - name: vol-mariadb
           persistentVolumeClaim:
             claimName: pvc-mariadb
+        - name: vol-my-conf
+          configMap:
+            name: cm-my-conf
+            items:
+              - key: etc-mysql-my.cnf
+                path: my.cnf
+        - name: vol-mariadb-conf
+          configMap:
+            name: cm-mariadb-conf
+            items:
+              - key: etc-mysql-mariadb.cnf
+                path: mariadb.cnf
+        - name: vol-mysql-safe-conf
+          configMap:
+            name: cm-mysql-safe-conf
+            items:
+              - key: etc-mysql-conf.d-mysqld_safe_syslog.cnf
+                path: mysqld_safe_syslog.cnf
--- a/k8s/kube-base/deployment-mariadb.yaml
+++ b/k8s/kube-base/deployment-mariadb.yaml
@@ -25,34 +25,19 @@ spec:
           volumeMounts:
             - name: vol-mariadb
               mountPath: /var/lib/mysql
-            - name: vol-my-conf
+            - name: vol-mariadb-conf
               mountPath: /etc/mysql/my.cnf
               subPath: my.cnf
             - name: vol-mariadb-conf
               mountPath: /etc/mysql/mariadb.cnf
               subPath: mariadb.cnf
-            - name: vol-mysql-safe-conf
+            - name: vol-mariadb-conf
               mountPath: /etc/mysql/conf.d/mysqld_safe_syslog.cnf
               subPath: mysqld_safe_syslog.cnf
       volumes:
         - name: vol-mariadb
           persistentVolumeClaim:
             claimName: pvc-mariadb
-        - name: vol-my-conf
-          configMap:
-            name: cm-my-conf
-            items:
-              - key: etc-mysql-my.cnf
-                path: my.cnf
         - name: vol-mariadb-conf
           configMap:
             name: cm-mariadb-conf
-            items:
-              - key: etc-mysql-mariadb.cnf
-                path: mariadb.cnf
-        - name: vol-mysql-safe-conf
-          configMap:
-            name: cm-mysql-safe-conf
-            items:
-              - key: etc-mysql-conf.d-mysqld_safe_syslog.cnf
-                path: mysqld_safe_syslog.cnf
--- a/k8s/kube-base/kustomization.yaml
+++ b/k8s/kube-base/kustomization.yaml
@@ -23,15 +23,11 @@ configMapGenerator:
   - name: cm-httpd-conf
     files:
       - httpd.conf
-  - name: cm-my-conf
-    files:
-      - mariadb-conf/etc-mysql-my.cnf
   - name: cm-mariadb-conf
     files:
-      - mariadb-conf/etc-mysql-mariadb.cnf
-  - name: cm-mysql-safe-conf
-    files:
-      - mariadb-conf/etc-mysql-conf.d-mysqld_safe_syslog.cnf
+      - my.cnf=mariadb-conf/etc-mysql-my.cnf
+      - mariadb.cnf=mariadb-conf/etc-mysql-mariadb.cnf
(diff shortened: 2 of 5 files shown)


k8s/kube-base/kustomization.yaml
```diff
@@ -23,10 +23,18 @@ configMapGenerator:
   - name: cm-httpd-conf
     files:
       - httpd.conf
+  - name: cm-my-conf
```
larsks, line 26: I would put all three files in a single configmap, rather than splitting them up like this.

k8s/kube-base/deployment-mariadb.yaml
```diff
@@ -25,7 +25,34 @@ spec:
           volumeMounts:
             - name: vol-mariadb
               mountPath: /var/lib/mysql
+            - name: vol-my-conf
+              mountPath: /etc/mysql/my.cnf
```
larsks, line 29: In the mariadb image, `/etc/mysql/my.cnf` is a symlink (to `/etc/mysql/mariadb.cnf`), so doesn't need to be replaced.

```
root@807f537634eb:/etc/mysql# ls -l /etc/mysql/my.cnf
lrwxrwxrwx. 1 root root 24 Aug  2 20:21 /etc/mysql/my.cnf -> /etc/alternatives/my.cnf
root@807f537634eb:/etc/mysql# ls -l /etc/alternatives/my.cnf
lrwxrwxrwx. 1 root root 22 Aug  2 20:21 /etc/alternatives/my.cnf -> /etc/mysql/mariadb.cnf
root@807f537634eb:/etc/mysql#

File: k8s/kube-base/mariadb-conf/etc-mysql-conf.d-mysqld_safe_syslog.cnf [hunk 3]
```diff
@@ -0,0 +1,4 @@
+# from configmap
```
larsks, line 1: Do you  want to add a similar comment to the other config files?

larsks (changes requested):

k8s/kube-base/deployment-mariadb.yaml
larsks: Hold on, I might be lying about this. Let me make sure I'm looking at the same image.

larsks: Never mind, this is a difference between the `mariadb:10` image (which gets us 10.8.3) and the explicit  10.3.34 you're using in the config.

k8s/kube-base/kustomization.yaml
```diff
@@ -23,10 +23,18 @@ configMapGenerator:
...
+    files:
+      - mariadb-conf/etc-mysql-my.cnf
+  - name: cm-mariadb-conf
+    files:
+      - mariadb-conf/etc-mysql-mariadb.cnf
+  - name: cm-mysql-safe-conf
+    files:
+      - mariadb-conf/etc-mysql-conf.d-mysqld_safe_syslog.cnf
 generatorOptions:
   disableNameSuffixHash: true
```
larsks, line 36: In general you don't want to do this: the suffix hash will allow openshift to automatically trigger a redeploy when you push an updated configmap (which otherwise won't take effect until you manually restart the relevant pod).

larsks:
The changes I'm suggesting resulting the `kustomization.yaml` looking like this:

```
commonLabels:
  app: xdmod
namespace: xdmod

resources:
  - pvc-mariadb.yaml
  - pvc-xdmod-conf.yaml
  - pvc-xdmod-src.yaml
  - pvc-xdmod-data.yaml
  - pvc-xdmod-openshift-data.yaml
  - pvc-clouds-yaml.yaml
  - pvc-httpd-conf.yaml
  - svc-mariadb.yaml
  - svc-xdmod.yaml
  - deployment-mariadb.yaml
  - deployment-xdmod.yaml
  - route-xdmod.yaml

configMapGenerator:
  - name: cm-xdmod-init-json
    files:
      - xdmod_init.json
  - name: cm-httpd-conf
    files:
      - httpd.conf
  - name: cm-mariadb-conf
    files:
      - my.cnf=mariadb-conf/etc-mysql-my.cnf
      - mariadb.cnf=mariadb-conf/etc-mysql-mariadb.cnf
      - mysqld_safe_syslog.cnf=mariadb-conf/etc-mysql-conf.d-mysqld_safe_syslog.cnf
generatorOptions:
  disableNameSuffixHash: true
  labels:
    type: generated
  annotations:
    note: generated
```

And `deployment-mariadb.yaml` looking like this:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mariadb
spec:
  selector:
    matchLabels:
      app: mariadb
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        app: mariadb
    spec:
      containers:
        - image: mariadb:10.3.34
          name: mariadb
          env:
            - name: MARIADB_ROOT_PASSWORD
              value: pass
          ports:
            - containerPort: 3306
              name: mariadb
          volumeMounts:
            - name: vol-mariadb
              mountPath: /var/lib/mysql
            - name: vol-mariadb-conf
              mountPath: /etc/mysql/my.cnf
              subPath: my.cnf
            - name: vol-mariadb-conf
              mountPath: /etc/mysql/mariadb.cnf
              subPath: mariadb.cnf
            - name: vol-mariadb-conf
              mountPath: /etc/mysql/conf.d/mysqld_safe_syslog.cnf
              subPath: mysqld_safe_syslog.cnf
      volumes:
        - name: vol-mariadb
          persistentVolumeClaim:
            claimName: pvc-mariadb
        - name: vol-mariadb-conf
          configMap:
            name: cm-mariadb-conf
```

k8s/kube-base/deployment-mariadb.yaml
rob-baron: Not certain which container you are looking at here.  For example, when I deploy from the ArgoCD set of yaml files, running your commands, I get:

        $ ls -la /etc/mysql/my.cnf
        -rw-r--r--. 1 root root 5221 Apr 30 01:20 /etc/mysql/my.cnf
        $ ls -la /etc/mysql/mariadb.cnf 
        -rw-r--r--. 1 root root 527 Feb 10  2022 /etc/mysql/mariadb.cnf

When gives 2 different files.  Furthermore, the my.cnf file that the following that loads mariadb.cnf with the following:

        #
        # * IMPORTANT: Additional settings that can override those from this file!
        #   The files must end with '.cnf', otherwise they'll be ignored.
        #
        !include /etc/mysql/mariadb.cnf

And loads more with the following (at the end of the my.cnf file):

        # Import all .cnf files from configuration directory
        !includedir /etc/mysql/mariadb.conf.d/
        
Perhaps there is some way that we can reconcile what you are seeing with what I am seeing?

k8s/kube-base/kustomization.yaml
rob-baron: I started doing it with 1 configmap, but it was easier to think about as 3 different configmaps.

k8s/kube-base/mariadb-conf/etc-mysql-conf.d-mysqld_safe_syslog.cnf
rob-baron: There were similar comments, though they were buried.  It is easier to see if it is the first line.

k8s/kube-base/kustomization.yaml
rob-baron: In this particular case, I found it useful.

for example, If I want to deploy something that relies on the naming of a configmap separately, I don't have to constantly change the suffix in my yaml.  I had initially tried working with out disabling the name suffix hash, but that was just difficult.

I'm happy to create the configmaps manually so that we don't have to disable the name suffix hash.

k8s/kube-base/kustomization.yaml
rob-baron: It's too bad you couldn't place you comment below, here, as it really deserves to be in this conversation as opposed to somewhere else.

It just makes this seem disjointed.

Ok, fine, I put all three files in a single confmap

joachimweyl:
@larsks can this be re-reviewed. @naved001 @knikolla please take a look.

larsks (approved):

PR #74, opened 2022-09-12
Fixes the connection timeout issue

rob-baron:
This change fixes the connection timeout issue in the manner as described by @larsks.

To summarize, kustomize overwrite the app label, so there needs to be a second label to allow the 2 services to connect to the appropriate pod.


larsks (approved):

k8s/kube-base/deployment-xdmod.yaml
```diff
@@ -1,20 +1,22 @@
...
 metadata:
-  name: moc-xdmod
+  name: xdmod-ui
+  labels:
+    component: xdmod-ui
 spec:
   selector:
     matchLabels:
-      app: xdmod
+      component: xdmod-ui
```
naved001, line 10: Sorry, I am a little confused. You deleted `k8s/overlays/nerc-ocp-infra/patches/deployment-xdmod.yaml`, which would add the label `app: xdmod` to the deployment named `moc-xdmod`. But now that the deployment name is changed to xdmod-ui and there's no patch in the overlay, how will the app label be added here?  Or do you just not plan to add it.

naved001: Also, that patch was setting some container images too, is that not needed anymore?

k8s/kube-base/deployment-xdmod.yaml
```diff
@@ -63,7 +65,7 @@ spec:
               mountPath: /root/httpd
             - name: vol-xdmod-init
               mountPath: /root/xdmod_init
-        - image: moc-xdmod-dev
```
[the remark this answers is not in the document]
rob-baron: Yes it is.  moc-xdmod-dev has high enough privileges to read from the directories that it needs to read from to initialize the persistent volumes.  Once the persistent volumes are initialized, the other containers don't need elevated permissions to read and write to the mounted volumes.

k8s/kube-base/deployment-xdmod.yaml
rob-baron: Not quite how it seems to work.

app: xdmod is being added by the k8s/kube-base/kustomization file though the commonLabels section:

        commonLabels:
          app: xdmod

I decided to change the name of the deployment to match the component as it is more descriptive of what it is deploying.  That is the xdmod-ui.  moc-xdmod doesn't really seem to fit any more, as if it ever did, and xdmod is built of several different pieces so I am trying to more descriptive of what each piece is.

k8s/overlays/nerc-ocp-infra/patches/deployment-xdmod.yaml was not being referred to in the k8s/overlays/nerc-ocp-infra/kustomization fils:

        kind: Kustomization
        resources:
          - ../../xdmod-build
          - ../../kube-base
        patchesStrategicMerge:
          - patches/bc-xdmod.yaml
          - patches/bc-xdmod-dev.yaml

So why keep it?

naved001: @rob-baron thank you, that makes sense.

naved001 (approved):

PR #81, opened 2022-09-26
Remove Trigger

rob-baron:
Removed this trigger as it doesn't currently work.


naved001 (approved):

knikolla (approved):
A quick google search shows that `triggers` is an attribute of `DeploymentConfig` rather than `Deployment`.

PR #84, opened 2022-11-10
Compute volume events

rob-baron:
This fixes issues discovered shredding and ingesting the events being generated by this script.

related to  issue (https://github.com/CCI-MOC/xdmod-cntr/issues/82)

### DIFF
--- a/moc_openstack_api_reporting.py
+++ b/moc_openstack_api_reporting.py
@@ -361,42 +361,62 @@ def compile_server_state(server, project_dict, flavor_dict, user_dict):
     if server.user_id in user_dict:
         user_name = user_dict[server.user_id]["name"]
 
+    # because the NERC has flavors IDs that don't exist in their flavor dictionary
+    disk_gb = "0"
+    ephemeral_gb = "0"
+    instance_type = f"unknown flavor({server.flavor['id']})"
+    memory_mb = "0"
+    root_gb = "0"
+    vcpus = "0"
+    if server.flavor["id"] in flavor_dict:
+        disk_gb = str(flavor_dict[server.flavor["id"]]["disk"])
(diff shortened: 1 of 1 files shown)


moc_openstack_api_reporting.py
```diff
@@ -361,42 +361,62 @@ def compile_server_state(server, project_dict, flavor_dict, user_dict):
...
+    root_gb = "0"
+    vcpus = "0"
+    if server.flavor["id"] in flavor_dict:
+        disk_gb = str(flavor_dict[server.flavor["id"]]["disk"])
+        ephemeral_gb = str(flavor_dict[server.flavor["id"]]["ephemeral"])
+        instance_type = flavor_dict[server.flavor["id"]]["name"]
+        memory_mb = str(flavor_dict[server.flavor["id"]]["ram"])
+        root_gb = str(flavor_dict[server.flavor["id"]]["disk"])
+        vcpus = str(flavor_dict[server.flavor["id"]]["vcpus"])
+
```
larsks, line 378: This works, but if you're interested it's possible to substantially reduce the amount of code you're introducing here.

You could instead do something like:

```python
from collections import defaultdict

flavor_dict.setdefault(server.flavor["id"],
  defaultdict(lambda: "0", name=f"unknown flavor({server.flavor['id']})"))
```

If `flavor_dict[server.flavor["id"]]` doesn't exist, this sets it to a dictionary that returns `"0"` for every key except for `name`.

This allows the following code to remain unmodified; code like this will work properly even if the flavor was missing:

```
"ephemeral_gb": flavor_dict[server.flavor["id"]]["ephemeral"],
```

larsks (approved):

naved001 (approved):

PR #89, opened 2022-12-05
Add missing cron jobs to kustomization.yaml

tzumainn:
kustomization.yaml was missing the cron jobs expected to run; this PR adds them in (as well as renaming the cron job yaml files to be prefixed with 'cron-job').


tzumainn:
@rob-baron @larsks let me know what you think!

larsks (approved):
Looks good to me!

Issue #90, opened 2022-12-05
Correct validation errors in cronjob manifests


larsks:
The recently added CronJob manifests are failing to apply:

```
$ kustomize build | kubectl apply -f- --dry-run=client
namespace/xdmod configured (dry run)
configmap/cm-httpd-conf configured (dry run)
configmap/cm-mariadb-conf configured (dry run)
configmap/cm-xdmod-init-json configured (dry run)
service/mariadb configured (dry run)
service/xdmod configured (dry run)
persistentvolumeclaim/pvc-clouds-yaml configured (dry run)
persistentvolumeclaim/pvc-httpd-conf configured (dry run)
persistentvolumeclaim/pvc-mariadb configured (dry run)
persistentvolumeclaim/pvc-xdmod-conf configured (dry run)
persistentvolumeclaim/pvc-xdmod-data configured (dry run)
persistentvolumeclaim/pvc-xdmod-openshift-data configured (dry run)
persistentvolumeclaim/pvc-xdmod-src configured (dry run)
deployment.apps/mariadb configured (dry run)
deployment.apps/xdmod-ui configured (dry run)
error: error validating "STDIN": error validating data: ValidationError(CronJob.spec.jobTemplate.spec.template.spec): unknown field "backoffLimit" in io.k8s.api.core.v1.PodSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

The `backoffLimit` field is applied in different places in each resource (and in both cases it is being applied in the wrong place).

The `backoffLimit` field is a member of `CronJob.spec.jobTemplate.spec`.

larsks:
@rob-baron @tzumainn  there are some validation errors in those cron jobs.

PR #91, opened 2022-12-05
Correct validation errors in CronJobs

larsks:
The `backoffLimit` field was in the wrong place on both of the recently
added CronJob resources. This commit corrects than manifests to use
CronJob.spec.jobTemplate.spec.backoffLimit.

Closes: #90


tzumainn (approved):
I've verified that the updated yaml will pass validation. Thanks for catching this!

rob-baron (approved):
Good catch larsks.

PR #93, opened 2022-12-06
Removed unused volumeMounts in cron-job-xdmod-openshift-prod.yaml

tzumainn:
The XDMoD OpenShift prod job had some leftover volumeMounts that are not needed. This PR removes them.


tzumainn:
@rob-baron @larsks Apologies for the noise, but this should fix the issues with this job (I double-checked with a server-side dry run).

larsks (approved):

rob-baron (approved):

PR #97, opened 2022-12-07
Update imagePullPolicy to Always

tzumainn:
Since we are pulling the latest version of the image, it makes sense to use 'Always' as the imagePullPolicy as it allows these jobs to immediately grab updated images.


tzumainn:
@rob-baron let me know what you think! The genesis of this patch is that the xdmod openshift jobs still aren't pulling the correct metric, despite the updated image; I think this is the reason why.

rob-baron:
I'm approving this as it should be fine in this environment.

PR #99, opened 2022-12-08
Adds missing copy of xdmod_init.json

tzumainn:
The xdmod openshift script was missing xdmod_init.json, as it needs to be copied over. This PR fixes that.

### DIFF
--- a/k8s/kube-base/xdmod_init.json
+++ b/k8s/kube-base/xdmod_init.json
@@ -42,6 +42,11 @@
             "name": "nerc_openshift_staging",
             "formal_name": "NERC OpenShift Staging",
             "type": "hpc"
+        },
+        {
+            "name": "nerc_openshift_prod",
+            "formal_name": "NERC OpenShift Prod",
+            "type": "hpc"
         }
     ]
 }
--- a/xdmod-openshift-reporting.sh
+++ b/xdmod-openshift-reporting.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+cp /mnt/xdmod_init/xdmod_init.json /etc/xdmod/xdmod_init.json
 /usr/bin/xdmod-get-config-files 
 /usr/bin/sleep 30
 
--- a/k8s/kube-base/cron-job-xdmod-openshift-prod.yaml
+++ b/k8s/kube-base/cron-job-xdmod-openshift-prod.yaml
@@ -13,7 +13,7 @@ spec:
           containers:
             - name: xdmod-openshift-prod-job
               image: image-registry.openshift-image-registry.svc:5000/xdmod/moc-xdmod:latest
-              imagePullPolicy: IfNotPresent
+              imagePullPolicy: Always
               envFrom:
                 - secretRef:
                     name: xdmod-openshift-prod-secret


tzumainn:
@rob-baron this should fix an issue with the xdmod_init.json file being missing for the openshift job. I've turned the image pull policy back to Always so that this change will pull in an updated image once it's rebuilt, but let me know if there's a better way.

tzumainn:
Also, I'm not certain - will updating xdmod_init.json be enough to create the new nerc_openshift_prod resource in xdmod? Or is there another step?

k8s/kube-base/xdmod_init.json
```diff
@@ -42,6 +42,11 @@
             "name": "nerc_openshift_staging",
             "formal_name": "NERC OpenShift Staging",
             "type": "hpc"
+        },
+        {
+            "name": "nerc_openshift_prod",
+            "formal_name": "NERC OpenShift Prod",
+            "type": "hpc"
         }
```
rob-baron, line 50: This creates a resource in /etc/xdmod/resource.json.  It controls what the resource is called in xdmod.

when this changes, we need to manually get the changes through by running xdmod-init, or by directly running xdmod-setup.

I'm fine with this now, though I would prefer to have staging setup the same as production.

xdmod-openshift-reporting.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+cp /mnt/xdmod_init/xdmod_init.json /etc/xdmod/xdmod_init.json
 /usr/bin/xdmod-get-config-files
```
rob-baron, line 4: you don't need to do this - in fact is it counter productive.

All of the config files, including xdmod_init.json are copied by one of the xdmod-ui init containers to PV that is then uploaded to the database.  It is this directory that is downloaded by xdmod-get-config-files so everything in the /etc/xdmod directory is overwritten.  I'm not quite certain how it will behave as we mount an ephemeral volume at that location and then you are copying the xdmod_init.json file there.

Besides, it is not where xdmod-init/xdmod-setup should be run from.  Personally, I consider xdmod-setup a little dangerous due to it failing in the middle of operations.

xdmod-openshift-reporting.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+cp /mnt/xdmod_init/xdmod_init.json /etc/xdmod/xdmod_init.json
 /usr/bin/xdmod-get-config-files 
 /usr/bin/sleep 30
```
rob-baron, line 5: you can probably remove this sleep.

rob-baron (changes requested):

xdmod-openshift-reporting.sh
tzumainn: I actually tested this - /etc/xdmod/xdmod_init.json does not exist in the running job. Exec'ing this command allowed it to work and connect to the database.

I copied this command from https://github.com/CCI-MOC/xdmod-cntr/blob/main/xdmod-run-ingestor.sh#L3 ; I suspect the ingestor job won't work without it as well... ?

k8s/kube-base/xdmod_init.json
tzumainn: Well, we can always redeploy with this change in staging after it's merged, correct?

xdmod-openshift-reporting.sh
tzumainn: So I tested re-running xdmod-get-config-files on an xdmod openshift prod job that was looping trying to connect to the database:

```
[stack@localhost ~]$ oc exec xdmod-openshift-prod-job-27842460-nsf76 /usr/bin/xdmod-get-config-files --as system:admin
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
/usr/local/lib/python3.6/site-packages/mysql/connector/authentication.py:44: CryptographyDeprecationWarning: Python 3.6 is no longer supported by the Python core team. Therefore, support for it is deprecated in cryptography and will be removed in a future release.
  from cryptography.exceptions import UnsupportedAlgorithm
Ensure the xdmod-init.json file is in /etc/xdmod/xdmod_init.json
Traceback (most recent call last):
  File "/usr/bin/xdmod-get-config-files", line 72, in <module>
    main()
  File "/usr/bin/xdmod-get-config-files", line 56, in main
    cnx = connect_to_db(config)
  File "/usr/bin/xdmod-get-config-files", line 22, in connect_to_db
    host = database["host"]
KeyError: 'host'
command terminated with exit code 1

```

However, if I copy the xdmod_init.json file first, it works:

```
[stack@localhost ~]$ oc exec xdmod-openshift-prod-job-27842460-nsf76 cp /mnt/xdmod_init/xdmod_init.json /etc/xdmod/.  --as system:admin
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
[stack@localhost ~]$ oc exec xdmod-openshift-prod-job-27842460-nsf76 /usr/bin/xdmod-get-config-files --as system:admin
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
<snip>

```

And after I run the latter, the job completes (and only ultimately fails because the nerc_openshift_prod resource doesn't exist in xdmod).

So in my view, xdmod_init.json is required; is there an example of xdmod-get-config-files running and working without it?

k8s/kube-base/xdmod_init.json
rob-baron: That's why I am fine with it.

xdmod-openshift-reporting.sh
rob-baron: This is probably what I did with the ingestor as I needed the credential and the credential is not in portal_settings.ini file yet.  I know it works there and I would prefer this to work the same way,  But I can do that after we deploy everything.

rob-baron (approved):

xdmod-openshift-reporting.sh
rob-baron: Sorry, I was being dense, I though that you had copied the script from the ingestor, however I now understand that you missed the first copy.

tzumainn: No worries, xdmod is super weird. Thanks for merging!

PR #107, opened 2022-12-22
Built xdmod open stack

rob-baron:
This for building the xdmod-openstack container using a build config from and a docker file that uses python 3.9 as it's base.  Major difference being that xdmod is based on centos7 and python 3.9 is based on ubuntu.

### DIFF
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -25,7 +25,7 @@ RUN yum makecache && \
     libreoffice \
     mariadb-server mariadb cronie logrotate \
     perl-Image-ExifTool php-mbstring php-pecl-apcu jq \
-    chromium-headless librsvg2-toolsm 
+    chromium-headless librsvg2-tools
 
 RUN yum clean all && \
     rm -rf /var/cache/yum
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -25,7 +25,16 @@ RUN yum makecache && \
     libreoffice \
     mariadb-server mariadb cronie logrotate \
     perl-Image-ExifTool php-mbstring php-pecl-apcu jq \
-    chromium-headless librsvg2-tools
+    chromium-headless librsvg2-toolsm \
+    gcc openssl-devel bzip2-devel libffi-devel
+
+RUN mkdir /python-src && \
+    cd /python-src && \
+    wget https://www.python.org/ftp/python/3.8.1/Python-3.8.1.tgz && \
+    tar -xzf Python-3.8.1.tgz && \
+    cd Python-3.8.1 && \
+    ./configure --enable-optimizations && \
+    make altinstall
 
 RUN yum clean all && \
     rm -rf /var/cache/yum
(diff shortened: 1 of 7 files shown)


Dockerfile.xdmod-openstack
```diff
@@ -0,0 +1,45 @@
...
+    && chmod 774 /app/run-xdmod-openstack.sh \
+    && sed -i -e 's/\r$//' /app/run-xdmod-openstack.sh \ 
+    && chmod 774 /app/run-xdmod-openstack-hypervisor.sh \
+    && sed -i -e 's/\r$//' /app/run-xdmod-openstack-hypervisor.sh \
+    && chmod -R g+rwx /app \
+    && chgrp -R 0 /app 
+
+# force downgrade of mysql-connector-python
+# https://stackoverflow.com/questions/73244027/character-set-utf8-unsupported-in-python-mysql-connector
+RUN pip3 install -I mysql-connector-python==8.0.29
```
larsks, line 39: What is this for? You've already installed `mysql-connector-python==8.0.29` on line 14, and there are no other `pip` operations in between that would have updated the installed pacakge.

Dockerfile.moc-xdmod
```diff
@@ -25,7 +25,16 @@ RUN yum makecache && \
     libreoffice \
     mariadb-server mariadb cronie logrotate \
     perl-Image-ExifTool php-mbstring php-pecl-apcu jq \
-    chromium-headless librsvg2-tools
+    chromium-headless librsvg2-toolsm \
+    gcc openssl-devel bzip2-devel libffi-devel
+
+RUN mkdir /python-src && \
```
larsks, line 31: The commit message for this change says, "Remove the source build for a newer Python version", but it appears to be **adding** code to build Python from source. I am confused.

rob-baron: I meant to remove it as it is simpler and less invasive to run this in a python container.

Dockerfile.xdmod-openstack
rob-baron: Will remove this.  This was just copied from the other docker file.

naved001:
The first commit is already in the repository (#105), might want to pull in the new changes (or rebase). There are some merge conflicts, so you'll have to do that anyway.

naved001 (approved):

Dockerfile.moc-xdmod
```diff
@@ -25,7 +25,7 @@ RUN yum makecache && \
     libreoffice \
     mariadb-server mariadb cronie logrotate \
     perl-Image-ExifTool php-mbstring php-pecl-apcu jq \
-    chromium-headless librsvg2-tools
+    chromium-headless librsvg2-toolsm
```
larsks, line 28: I think this is a typo: there is a package `librsvg2-tools`, but there is no package `librsvg2-toolsm`.

k8s/xdmod-build/bc-xdmod-openstack.yaml
```diff
@@ -0,0 +1,24 @@
+piVersion: build.openshift.io/v1
```
larsks, line 1: Typo: you're missing the `a` in `apiVersion`.

larsks (changes requested):

larsks (approved):

PR #109, opened 2023-01-02
Deploy Shred Job

rob-baron:
These are the files to deploy the openstack shredding job. 1) the configuration is loaded from the database via the init container 2) The data from openstack is loaded into the PV in the init container 3) xdmod-shredder is run in the container.

This was done as we don't yet support RWX persistent volumes and the xdmod team is still on centos 7 which uses an older version of of python of which they are planning to update to a current version of python.

I would like to patch the args to the command in the overlay/nerc-ocp-infra patch directory, but seem to be hitting https://github.com/kubernetes-sigs/kustomize/issues/4062

### DIFF
--- a/k8s/kube-base/cj-xdmod-openstack-shred.yaml
+++ b/k8s/kube-base/cj-xdmod-openstack-shred.yaml
@@ -9,12 +9,47 @@ spec:
     spec:
       template:
         spec:
+          initContainers:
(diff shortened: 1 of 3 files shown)


rob-baron:
This has been running on ocp-stagging without any issues, and I will push it late this evening or tomorrow morning to finish the configuration on the nerc cluster (assuming that people are not yet back to work).

PR #119, opened 2023-01-13
Qualifying the local registry for the moc-xdmod* images

rob-baron:
This is to test a work-a-round for the following issue:

    https://github.com/OCP-on-NERC/operations/issues/80

I have occasionally had to do this in the past and I am not certain why I didn't have to do this before the cluster rebuild.  I do prefer to not qualify the local registry.


rob-baron:
Note, I will push this later this evening, as I don't expect anyone to be working until tuesday, and I want to get xdmod up and running by then.

PR #120, opened 2023-01-17
Added the entire deployment to the patch

rob-baron:
As I am getting the following in the log files of the first init container (xdmod-init-1)  (currently that system is being reinstalled):

  (13)Permission denied: AH00072: make_sock: could not bind to address [::]:80
  (13)Permission denied: AH00072: make_sock: could not bind to address 0.0.0.0:80
  no listening sockets available, shutting down
  AH00015: Unable to open logs

This implies 2 things:

  1) kustomize is forgetting or patching over the command in the initial deployment
     as it is using the command for the docker conatiner as opposed to the one specified in
     the deployment.

  2) kustomize is probably patching over the entire container spec as it is not mounting
     the other directories, as an emphemerial disk is mounted to the log directory making
     it writeable.

As such, I am including the entire defintiion for all of the containers in the patch.

### DIFF
--- a/k8s/overlays/nerc-ocp-infra/patches/deployment-xdmod.yaml
+++ b/k8s/overlays/nerc-ocp-infra/patches/deployment-xdmod.yaml
@@ -8,6 +8,8 @@ spec:
   selector:
     matchLabels:
       component: xdmod-ui
+  strategy:
+    type: Recreate
   template:
     metadata:
       labels:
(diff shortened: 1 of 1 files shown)


larsks (changes requested):
Are you suggesting that there is a bug in kustomize, or are you suggesting that there was an error in the previous patch? It's not clear from the description here. If the former, we should put together a reproducer and create an issue for kustomize.

larsks:
This patch does not appear to impact the generated output:

```
$ git checkout main
$ kustomize build k8s/overlays/nerc-ocp-infra/ > main.yaml
$ gh pr checkout 120
Switched to branch 'AddCmdToPatch'
Your branch is up to date with 'upstream/AddCmdToPatch'.
Already up to date.
$ kustomize build k8s/overlays/nerc-ocp-infra/ > pullrequest.yaml
$ diff main.yaml pullrequest.yaml
<no output>
```

Because this patch is effectively a no-op I don't think we should merge it.

rob-baron:
@larsks, I agree, it would seem that the nerc-ocp-infra install went a little haywire.  Don't quite understand why the log file has any apache warnings/errors in it as the first init file doesn't run apache.  Furthermore, viewing the output shows both the correct command and the mount volumes.

I will just close this.

PR #121, opened 2023-01-19
Rework of the Ingestor job

rob-baron:
Split off the downloading of the config from the database into an init container from the running of xdmod-ingestor in the main container.


k8s/kube-base/cron-job-xdmod-ingestor.yaml
```diff
@@ -9,12 +9,27 @@ spec:
...
+                - name: vol-xdmod-init
+                  mountPath: /mnt/xdmod_init
+                - name: vol-var-log-xdmod
+                  mountPath: /var/log/xdmod
           containers:
             - name: moc-xdmod
               image: moc-xdmod
               imagePullPolicy: IfNotPresent
               command:
-                - "/usr/bin/xdmod-run-ingestor.sh"
```
knikolla, line 17: Since this file is not being used anymore it might be worth deleting it from the repo.

knikolla (approved):

k8s/kube-base/cron-job-xdmod-ingestor.yaml
rob-baron: Yup eliminated it.  You caught this after I created it, but before I moved it to draft.

Dockerfile.xdmod-openstack
```diff
@@ -14,7 +14,7 @@ RUN apt-get update \
 COPY ./hypervisor_facts.py /app/xdmod-openstack-hypervisor
 COPY ./moc_openstack_api_reporting.py /app/xdmod-openstack-reporting
 COPY ./GetConfigFiles.py /app/xdmod-get-config-files
-COPY ./xdmod-run-ingestor.sh /app/xdmod-run-ingestor.sh
+COPY ./xdmod-get-config.sh  /app/xdmod-get-config.sh
```
knikolla, line 17: This file is only copied over for the `xdmod-openstack` image. Is it not necessary for `moc-xdmod` image?

k8s/kube-base/cron-job-xdmod-ingestor.yaml
rob-baron: I haven't quite finished testing this and was planning to remove that file.  which is why I hadn't assigned any people to review.

Dockerfile.xdmod-openstack
rob-baron: Yes, I was thinking of removing it when I refactor the shredder

PR #123, opened 2023-01-23
Refactor shredder job

rob-baron:
This is a refactoring of the shredder job using a separate init container to download the configuration from the database as we currently do not have the facility to support RWX persistent volumes.

Furthermore, if and when we do get RWX persistent volumes, it should be an easy change to remove the init container.


rob-baron:
Note, this is still in progress (GitHub will not allow me to convert to drag as there are people subscribed to this).

rob-baron:
As I am leaving BU as of 06-SEP-2023, there is no need for this draft PR.

PR #124, opened 2023-01-25
Specify the local repo in order to find the image

rob-baron:
This way openstack will pull from the local repo as opposed to trying to pull from somewhere else and failing.  The containers found in this job are build locally from the build configs.

### DIFF
--- a/k8s/overlays/nerc-ocp-infra/kustomization.yaml
+++ b/k8s/overlays/nerc-ocp-infra/kustomization.yaml
@@ -10,4 +10,5 @@ resources:
 
 patchesStrategicMerge:
   - patches/cj-xdmod-openstack-shred.yaml
+  - patches/cj-xdmod-ingestor.yaml
   - patches/deployment-xdmod.yaml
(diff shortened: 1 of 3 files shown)


k8s/overlays/nerc-ocp-infra/kustomization.yaml
```diff
@@ -10,4 +10,5 @@ resources:
 
 patchesStrategicMerge:
   - patches/cj-xdmod-openstack-shred.yaml
+  - patches/cj-xdmod-ingestor.yaml
```
knikolla, line 13: This seems unrelated?

knikolla: Ah, had to refresh. Now I see it.

knikolla (approved):

k8s/overlays/nerc-ocp-infra/kustomization.yaml
rob-baron: how so?

PR #126, opened 2023-01-26
Specifies OpenShift Interanl Registry for the Init Container Image

rob-baron:
There is one more image that needs to be pulled from the local registry and is blocking the shredder job from running (as openshift is unable to pull the image)

Here is the error message from the logs that this is expected solve:

```
oc logs cj-xdmod-openstack-27911400-jfskx -c cj-init-xdmod-openstack
Error from server (BadRequest): container "cj-init-xdmod-openstack" in pod "cj-xdmod-openstack-27911400-jfskx" is waiting to start: trying and failing to pull image
```

Note, this was not occurring before the rebuilds on the infra cluster and is not currently needed on ocp-staging.


rob-baron:
I am just going to squash and merge this as I don't expect anyone to be around tonight and this was a simple oversight from the last PR.

PR #129, opened 2023-01-30
Remove google chrome from the moc-xdmod docker image

rob-baron:
As google chrome is only used to run their UI tests and we are not yet developing anything related to the UI so we don't need to refer to google chorme in the dockerfile.


knikolla (changes requested):
Please remove the repo as well, on line 4. And from the assets folder. 

EDITED: to also mention removing from assets folder.

rob-baron:
ok, the repo is removed from line 4

larsks (approved):

knikolla (approved):

Issue #130, opened 2023-01-31
Use before initialization error in moc_openstack_api_reporting.py


rob-baron:
As there are more messages that exist that are not handled by xdmod, sometime the event_data (line  ~533) is used before it is initialized.

PR #131, opened 2023-01-31
This Fixes a Use of event_data Before Initialization

rob-baron:
There are events that occur which xdmod doesn't handle and so event_data is not tracked.

https://github.com/CCI-MOC/xdmod-cntr/issues/130

### DIFF
--- a/moc_openstack_api_reporting.py
+++ b/moc_openstack_api_reporting.py
@@ -497,7 +497,7 @@ def process_compute_events(openstack_conn, script_datetime, openstack_data, clus
     last_run_datetime = datetime.datetime.fromisoformat(cluster_state["last_run_timestamp"])
 
     for server in openstack_data["server_dict"].values():
-        event_data_defined = 0
+        event_data = dict()
         # need to generate an existence event for each server in server_dict
         if server["state"] == "ACTIVE" or server["state"] == "DELETED":
             event_data = {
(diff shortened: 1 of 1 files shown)


moc_openstack_api_reporting.py
```diff
@@ -497,6 +497,7 @@ def process_compute_events(openstack_conn, script_datetime, openstack_data, clus
     last_run_datetime = datetime.datetime.fromisoformat(cluster_state["last_run_timestamp"])
 
     for server in openstack_data["server_dict"].values():
+        event_data_defined = 0
```
knikolla, line 500: Is there any reason in particular why this is an integer instead of a boolean? To me it feels like it should be a boolean `event_data_defined = True`. This makes the code easier to read and understand.

knikolla (changes requested):

moc_openstack_api_reporting.py
knikolla: Though in general, a better approach would be to define `event_data = dict()` before you start using it, as an empty dict. And then check for `if event_data:`. The conditional will return false for an empty_dict.

knikolla (approved):
Thank you!

Also, if you update the PR description to say Closes #<number of the issue>, then it will be automatically closed when the PR is approved. Just writing the issue number doesn't automatically perform the issue closing.

PR #132, opened 2023-02-01
Fixes undefined dictionary key

rob-baron:
Referenced to a non-existant dictionary item.

### DIFF
--- a/moc_openstack_api_reporting.py
+++ b/moc_openstack_api_reporting.py
@@ -734,7 +734,7 @@ def main():
 
     vol_keys = list(cluster_state["vol_timestamps"].keys())
     for key in vol_keys:
-        if cluster_state["vol_timestamps"][key]["updated"] == 0:
+        if "updated" in list(cluster_state["vol_timestamps"][key].keys()) and cluster_state["vol_timestamps"][key]["updated"] == 0:
             del cluster_state["vol_timestamps"][key]
         else:
             del cluster_state["vol_timestamps"][key]["updated"]
--- a/moc_openstack_api_reporting.py
+++ b/moc_openstack_api_reporting.py
@@ -734,7 +734,7 @@ def main():
 
     vol_keys = list(cluster_state["vol_timestamps"].keys())
     for key in vol_keys:
-        if "updated" in list(cluster_state["vol_timestamps"][key].keys()) and cluster_state["vol_timestamps"][key]["updated"] == 0:
+        if cluster_state["vol_timestamps"][key].get("updated") == 0:
             del cluster_state["vol_timestamps"][key]
         else:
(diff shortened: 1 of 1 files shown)


moc_openstack_api_reporting.py
```diff
@@ -734,7 +734,7 @@ def main():
 
     vol_keys = list(cluster_state["vol_timestamps"].keys())
     for key in vol_keys:
-        if cluster_state["vol_timestamps"][key]["updated"] == 0:
+        if "updated" in list(cluster_state["vol_timestamps"][key].keys()) and cluster_state["vol_timestamps"][key]["updated"] == 0:
```
knikolla, line 737: A better way to write this is `if cluster_state["vol_timestamps"][key].get("updated") == 0:`

The get method returns `None` if the dictionary doesn't contain the key, and None is different from 0.

naved001, line 737: You could use the get method on the dictionary.

`cluster_state["vol_timestamps"][key].get("updated") == 0`

because `cluster_state["vol_timestamps"][key].get("updated")` would return None when "updated" doesn't exist and the comparison would then correctlya fail.

knikolla (changes requested):

naved001 (approved):
Left a suggestion, don't feel strongly about it.

moc_openstack_api_reporting.py
```diff
@@ -734,7 +734,7 @@ def main():
 
     vol_keys = list(cluster_state["vol_timestamps"].keys())
     for key in vol_keys:
```
knikolla, line 736: This is unrelated to this PR and I'm not asking you to make this change. Just noting this as an opportunity to write better Python code in the future.

In Python 3, you can iterate through keys in a dictionary just by doing `for key in cluster_state["vol_timestamps"]`. Turning it into a list is not necessary.

naved001: I think it's because he's deleting the list he's iterating over, so that's why he wanted a copy of just the keys.

knikolla: Good point, thanks. I missed that.

rob-baron: @naved001 is absolutely correct.  Since we are deleting keys we need a copy that is not changing.

moc_openstack_api_reporting.py
rob-baron: sure

knikolla (approved):
Thanks for updating it.

When you write commit messages, or PR titles, it would be much more helpful if instead of writing "who" made the suggestion, you write what is it that you are doing. This way someone that is going through the commit, doesn't need to have the context of what "kristi's suggestion is", but can instead see something like "Fixed to use get() method instead of converting into a list of keys".

PR #133, opened 2023-02-01
Fix Bug 730 take 2

rob-baron:
This is a second take to fix the bug at line 730

### DIFF
--- a/moc_openstack_api_reporting.py
+++ b/moc_openstack_api_reporting.py
@@ -727,17 +727,19 @@ def main():
 
     vm_keys = list(cluster_state["vm_timestamps"].keys())
     for key in vm_keys:
-        if cluster_state["vm_timestamps"][key].get("updated") == 0:
-            del cluster_state["vm_timestamps"][key]
-        else:
-            del cluster_state["vm_timestamps"][key]["updated"]
+        if cluster_state["vm_timestamps"][key].get("updated") is not None: 
+            if cluster_state["vm_timestamps"][key]["updated"] == 0:
+                del cluster_state["vm_timestamps"][key]
+            else:
+                del cluster_state["vm_timestamps"][key]["updated"]
 
     vol_keys = list(cluster_state["vol_timestamps"].keys())
     for key in vol_keys:
-        if cluster_state["vol_timestamps"][key].get("updated") == 0:
-            del cluster_state["vol_timestamps"][key]
-        else:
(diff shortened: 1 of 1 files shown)


moc_openstack_api_reporting.py
```diff
@@ -727,17 +727,19 @@ def main():
...
+            else:
+                del cluster_state["vm_timestamps"][key]["updated"]
 
     vol_keys = list(cluster_state["vol_timestamps"].keys())
     for key in vol_keys:
-        if cluster_state["vol_timestamps"][key].get("updated") == 0:
-            del cluster_state["vol_timestamps"][key]
-        else:
-            del cluster_state["vol_timestamps"][key]["updated"]
+        if cluster_state["vol_timestamps"][key].get("updated") is not None:
```
knikolla, line 738: You can check if a key is present in a dictionary by doing `<key> in <dict>` for example. 

```
>>> a = {'a': 1, 'b': 2}
>>> 'c' in a
False
>>> 'a' in a
True
```

So this line can be rewritten for readability as `if "updated" in cluster_state["vol_timestamps"][key]`.

knikolla (changes requested):

knikolla:
And thank you for providing a meaningful commit messages as to why we want to delete those entries.

moc_openstack_api_reporting.py
```diff
@@ -727,17 +727,19 @@ def main():
...
+            else:
+                del cluster_state["vm_timestamps"][key]["updated"]
 
     vol_keys = list(cluster_state["vol_timestamps"].keys())
     for key in vol_keys:
-        if cluster_state["vol_timestamps"][key].get("updated") == 0:
-            del cluster_state["vol_timestamps"][key]
-        else:
-            del cluster_state["vol_timestamps"][key]["updated"]
+        if "update" in cluster_state["vol_timestamps"][key]:
```
knikolla, line 738: I believe this is a typo here as it should be `"updated"`

knikolla (changes requested):

moc_openstack_api_reporting.py
rob-baron: yes it is a typo.  fixed.

knikolla (approved):
Thank you. When you merge this, please squash and merge this to reduce it from 4 commits to 1.

PR #134, opened 2023-02-01
The resource has an '_' where as the cloud config has a '-'

rob-baron:
This fix should fix this error:

    "Failed to create shredder: No config found for 'nerc-openstack' in '/etc/xdmod/resources.json'"

The resrouce is refered to as nerc_openstack, where as the cloud config uses nerc-openstack.  Keep in mind:

1) the cloud config (as defined in /etc/openstack/clouds.yaml) is used in the init container 2) the resource (as defined in /etc/xdmod/resources.json)  is used by the shredder.

### DIFF
--- a/k8s/overlays/nerc-ocp-infra/patches/cj-xdmod-openstack-shred.yaml
+++ b/k8s/overlays/nerc-ocp-infra/patches/cj-xdmod-openstack-shred.yaml
@@ -29,5 +29,5 @@ spec:
                   "-d",
                   "/data",
                   "-r",
-                  "nerc-openstack",
+                  "nerc_openstack",
                 ]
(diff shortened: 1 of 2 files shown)


knikolla (approved):
Seems to be the case https://github.com/CCI-MOC/xdmod-cntr/blob/43ecb204b1a45c6d0d30ab317b16b0c1a3ea15a8/k8s/kube-base/xdmod_init.json#L31

knikolla:
Do you need to update this one too? In ocp-staging overlay https://github.com/CCI-MOC/xdmod-cntr/blob/1345b04135d525863a78aac24612166b223ce2dd/k8s/overlays/nerc-ocp-staging/patches/cj-xdmod-openstack-shred.yaml#L31

larsks (approved):

rob-baron:
I should update ocp-staging as well - just to be consistent.  Though it is deployed manually.

PR #135, opened 2023-02-02
Added workflow to validate manifests

knikolla:
Lifted from `OCP-on-NERC/nerc-ocp-config`, this is useful for finding errors that would prevent the manifests from being loaded.

Also fixed `minikube` and `CRC` overlays pointing to the wrong directory names or not existing files.

### DIFF
--- a/.github/workflows/validate-manifests.yaml
+++ b/.github/workflows/validate-manifests.yaml
@@ -1,0 +1,24 @@
+name: Kustomize build overlays
+
+on:
+  pull_request:
+    paths:
+      - '**.yaml'
+
+jobs:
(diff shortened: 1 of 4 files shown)


larsks (approved):

rob-baron:
Actually, I was planning on getting rid of the minicube and CRC overlays as I haven't used them for a while.  I figure it would be one less thing to maintain, and if you need it, it would probably be easier to start with the infra or staging overlay.

rob-baron (approved):

knikolla:
Can be removed in a follow-up patch. Merging this in the meanwhile.

PR #136, opened 2023-02-06
Add xdmod-openshift-prod secret for token required to access prod data

tzumainn:


tzumainn:
@larsks does this look right? I wasn't sure if I'm supposed to use the same ClusterSecretStore

larsks:
That looks fine. Because the infra cluster is an "admin only" cluster there is a single `ClusterSecretStore` that can be used globally for provisioning secrets.

(On the prod cluster, there are only per-namespace `SecretStores` in specific namespaces.)

k8s/kube-base/cron-job-xdmod-openshift-prod.yaml
```diff
@@ -28,6 +25,11 @@ spec:
                   value: /data/xdmod_openshift_data/nerc_openshift_prod
                 - name: XDMOD_RESOURCE
                   value: nerc_openshift_prod
+                - name: OPENSHIFT_TOKEN
+                  valueFrom:
+                    secretKeyRef:
+                      key: token
+                      name: xdmod-openshift-prod-secret
```
larsks, line 32: Why the change from `envFrom`? When you create the secret, you could just create it with an `OPENSHIFT_TOKEN` key. Either solution is fine; I'm just curious what motivated this particular change.

larsks (approved):

k8s/kube-base/cron-job-xdmod-openshift-prod.yaml
tzumainn: Oh uh it's because all the examples in the documentation did it this way :)

larsks:
@knikolla it looks like you were missing a script in a9eeeaafd006bf7c8c80f6dca72c18a782b19987.

rob-baron (approved):

tzumainn:
@rob-baron thanks! do I need to do anything else to have this merged?

PR #137, opened 2023-02-06
Update xdmod-openshift cron job with initContainer to grab config

tzumainn:


tzumainn:
@rob-baron the other option is to update GetConfigFiles.py to specify /usr/bin/python3, but I don't know if that affects your openstack jobs... ?

rob-baron (changes requested):
@tzumainn Can you try to do this in an  init container with the python3 image as done here: (https://github.com/CCI-MOC/xdmod-cntr/blob/main/k8s/kube-base/cron-job-xdmod-ingestor.yaml lines  - 12-26)?

This way we isolates the change to a separate container which can be pulled (if/when we get RWX persistent volumes).

tzumainn:
> @tzumainn Can you try to do this in an init container with the python3 image as done here: (https://github.com/CCI-MOC/xdmod-cntr/blob/main/k8s/kube-base/cron-job-xdmod-ingestor.yaml lines - 12-26)?
> 
> This way we isolates the change to a separate container which can be pulled (if/when we get RWX persistent volumes).

Thanks for the suggestion! I've updated the PR; does it look right to you?

rob-baron (approved):

tzumainn:
Thanks! Is this mergeable now?

PR #140, opened 2023-02-10
Add /etc/openstack dir to init container

tzumainn:
The /etc/openstack directory is expected by the xdmod-get-config.sh script


tzumainn:
@rob-baron let me know what you think!

PR #141, opened 2023-02-13
Removed extraneous commands from xdmod-openshift script

tzumainn:
These commands are no longer needed now that the config is fetched in an init container


tzumainn:
@rob-baron I've been having trouble figuring out how to test these changes efficiently in the infra cluster; let me know if you have a good workflow!

Anyway, after the last PR was merged the initcontainer now finishes successfully, but the job still can't reach the database. I think it's because the key config files are in /etc/xdmod, and I didn't mount it.

rob-baron (approved):

tzumainn:
@rob-baron thanks! can this be merged as well?

PR #145, opened 2023-02-14
Work-A-Round for Running xdmod's acl-config

rob-baron:
It seems that everytime the shredder run, the xdmod-ui interface gets stuck in the loading process.  This is either due to acl-config not running to completion on the shdredder pod, or acl-config modifying the config directory and scince the config directory is not shared RWX it doesn't updated on the xdmod-ui pod.

This will require more investigation as to what is occurng.  Meanwhile, this will run acl-config in the xdmod-ui pod every 7 minutes.  I'm trying to strike a balance between running acl-config too often (consumning too much database time) and keeping the UI from being stuck loading.

### DIFF
--- a/k8s/kube-base/deployment-xdmod.yaml
+++ b/k8s/kube-base/deployment-xdmod.yaml
@@ -45,10 +45,14 @@ spec:
         - image: moc-xdmod:latest
           name: xdmod-acl-work-a-round
           imagePullPolicy: IfNotPresent
-          command: ["/usr/bin/bash"]
-          args: ["-c", "while true; do /usr/bin/acl-config; sleep 420; done"]
-          ports:
-            - containerPort: 8080
+          command: 
+            - /bin/bash
+            - -c
+            - |
+              while true; do
+                /usr/bin/acl-config
+                sleep 420
+              done
           volumeMounts:
             - name: vol-xdmod-conf
               mountPath: /etc/xdmod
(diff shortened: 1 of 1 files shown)


k8s/kube-base/deployment-xdmod.yaml
```diff
@@ -42,6 +42,34 @@ spec:
               mountPath: /run/httpd
             - name: vol-php-session
               mountPath: /var/lib/php/session
+        - image: moc-xdmod:latest
+          name: xdmod-acl-work-a-round
+          imagePullPolicy: IfNotPresent
+          command: ["/usr/bin/bash"]
+          args: ["-c", "while true; do /usr/bin/acl-config; sleep 420; done"]
+          ports:
```
larsks, line 50: The `ports` section here might be problematic, since you've already indicated that the `xdmod` container is using port 8080 (and we're not exposing a network service in this container in any case).

k8s/kube-base/deployment-xdmod.yaml
```diff
@@ -42,6 +42,34 @@ spec:
               mountPath: /run/httpd
             - name: vol-php-session
               mountPath: /var/lib/php/session
+        - image: moc-xdmod:latest
+          name: xdmod-acl-work-a-round
+          imagePullPolicy: IfNotPresent
+          command: ["/usr/bin/bash"]
+          args: ["-c", "while true; do /usr/bin/acl-config; sleep 420; done"]
```
larsks, line 49: What's here is absolutely fine, but consider combining `command` and `args` and using a literal block to make the code easier to read:

```
command:
  - /bin/bash
  - -c
  - |
    while true; do
      /usr/bin/acl-config
      sleep 420
    done
```

larsks (approved):

Issue #149, opened 2023-03-16
job to Backup database


rob-baron:
create a job that backups the mariadb database to an s3 volume.

Issue #86, opened 2022-11-17
Backups for all of the xdmod persistent volumes


rob-baron:
As it is likely that OpenShift will occasionally need to be reinstalled, we need to back up persistent volumes so that we can bring xdmod back up more quickly.

rob-baron:
On the xdmod side a minimal backup would be (accessible through the xdmod-ui container):
```
  1) tar -zcvf etc-xdmod.tgz /etc/xdmod
  2) tar -zcvf usr-share-xdmod.tgz /usr/share/xdmod
  3.0) mysqldump -h mariadb -u root -p --databases  > xdmod-db-backup.sql
  3.1) gzip xdmod-db-backup.sql
  4) oc rsync these off of the cluster
```
When I restored from backup on the ocp-staging environment, I got the interface back with the data.

rob-baron:
This has the same problem with sharing our configuration files.

So the basic plan is to save any directory that needs to be backed up to the database, and then downloading a mysqldump of the database. 

Saving directories to the database can be done in parallel with either the shredding jobs or the xdmod-ui process.

Storing the mysqldump file can be done in a kubernetes cronjob.

rob-baron:
First part is to upload tarballs to the database.

Second part is to dump the database to a location for the backup.

PR #151, opened 2023-03-17
xdmod backup persistent volumes to db

rob-baron:
Since we do not have RWX and we have a shared database, it is easy to backup the config/data directories to the database.

The plan is to use a different job to dump the contents of the database to an S3 volume in a separate job.

See:
   (https://github.com/CCI-MOC/xdmod-cntr/issues/86)
   (https://github.com/CCI-MOC/xdmod-cntr/issues/149)

resolves ((https://github.com/CCI-MOC/xdmod-cntr/issues/86)


store_to_db.py
larsks, line 1: If the goal is ultimately to move the data to an object store, why are we bothering to store the files into the database first? Why not just go from the filesystem to an object store in a single step? We could use existing tooling for that (e.g., the minio client, or the aws cli) with substantially less complexity.

store_to_db.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def exec_fetchone(cursor, sql_stmt, params=None):
+    """executes the sql stmt and fetches the first one in the result list"""
+    cursor.execute(sql_stmt, params)
+    result = cursor.fetchone()
+    return result[0]
+
+
+def connect_to_db(database):
```
larsks, line 20: I think we already have implementations of `connect_to_db` and `write_file_to_db` elsewhere in this repository. Let's refactor them into a common module rather than re-writing them.

store_to_db.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/local/bin/python3
+""" Module to gzip and store a directory in the database """
+import os
+import datetime
+import time
+import json
+import re
+import argparse
+import tempfile
```
larsks, line 9: This import isn't used.

store_to_db.py
```diff
@@ -0,0 +1,117 @@
...
+
+def process_directories(dir_list):
+    """stores each directory in the list to the database"""
+    for dir_name in dir_list:
+        key_val = key_from_dir(dir_name)
+        print("Processing {dir_name},{key_val}")
+        write_directory_to_db(dir_name, key_val)
+
+
+def process_continueously(dir_list):
```
larsks, line 96: We should eliminate this function (and `sleep_till_midnight`) and just use a cronjob for periodic execution. That gives us the flexibility of changing schedules without needing to patch the code.

larsks (changes requested):

store_to_db.py
larsks: > and makes it more difficult to have the possibility that the configuration in the files is not the same as the configuration as the database.

If that's a concern then we have bigger problems, because this would be backing up files from the filesystem -- that is, the files that are actually in use. If there's a chance that those are different from what's stored in the database, then we have a critical problem that needs fixing.

store_to_db.py
rob-baron: This was done to back up the persistent volumes of the xdmod-ui at midnight.  

I would prefer to use a kubernetes cron job as it would give us the flexibility (it is why I added separate cron jobs for the ingestor and shredders as the ingestor doesn't need to run on the same schedule as the shredders).  However, the kubernetes cron job would operate in a separate pod from the xdmod-ui pod and would not have access to the persistent volumes there as we don't have volumes that are RWX.

Creating a cron job that backs up all of the PVs would be simpler if we had RWX volumes - or is there a way that I can mount the PVs from separate pods as read only (as opposed to creating RWO PVs)?

store_to_db.py
rob-baron: It is only a minor concern.

I am thinking that if the database goes down, then it might be possible that the configuration doesn't match the database.  I have yet to see this happen to this project, and mariadb seems to be quite stable.  However, I have seen similar things happen to other projects and yes, it is a pain to fix.

That said, if we are using the database to store all of the configuration and data files, then it is more difficult for this to happen.

store_to_db.py
rob-baron: @larsks - how about if I split the questions (explicit and implied) out:

1) If this were run as a kubernetes cron job, can it access the PVs that are attached to the pod running the UI without brining the UI down?

2) Would a command line parameter suffice, in lieu of a kubernetes cron job, as that would be editable from the deployment?

store_to_db.py
rob-baron: sure.

PR #153, opened 2023-03-19
feature/pre commit

larsks:
Add pre-commit configuration and workflows

Bring this repository in line with some of our development standards by adding a pre-commit configuration and matching GitHub workflow to enforce various linters.

We have restricted the Black python code formatter to version 3.8, because this is the version used in the xdmod image(s).

This pull request also includes all the changes generated by running the pre-commit hooks.


rob-baron (approved):

larsks:
> Most of the code has had black run against it. Is it the case that our standards as to the line width has decreased? Just noticing some lines that were run through black have been broken into multiple lines.

We're not explicitly configuring line lengths, so we're using the black defaults.

> Our standards for formatting the mysql config files make the standard config files from mariadb look a bit odd.

It looks like these are being caught by the "remote tabs" filter. We can exclude these files from the linters if you think that would be appropriate.

larsks:
I've updated the pull request to exclude `*.cnf` files from the tab remover.

rob-baron:
>>Most of the code has had black run against it. Is it the case that our standards as to the line width has decreased? Just >noticing some lines that were run through black have been broken into multiple lines.
>We're not explicitly configuring line lengths, so we're using the black defaults.

Sure, I thought it was ok to have long line lengths.  But this is fine as well.

>>Our standards for formatting the mysql config files make the standard config files from mariadb look a bit odd.
>It looks like these are being caught by the "remove tabs" filter. We can exclude these files from the linters if you think >that would be appropriate.

Honestly, it really doesn't matter to me (after all I have already approved it).  It is just when someone tries to compare the current files with the config files from mysql would have to realize that even though the formatting is different the most of the values have stayed the same.  I am unlikely to even bother comparing the original files, though I am acutely aware of people who would do so.

naved001 (approved):

PR #154, opened 2023-03-21
Add image build workflow

larsks:
The image is built for both pull requests and pushes.

The image is pushed to ghcr.io for:

- Pushes to main, which will be tagged with the branch name and the commit id.

- Pushes for semver [1] style tags (`vX.Y.Z`), which will be tagged with `vX.Y.Z`, `vX.Y`, `vX`, and `latest`.

This ensures that the Dockerfile builds successfully, makes the image available for deployment or local testing, and opens up the possibility of implementing some CI tests of the image in future work.

[1]: https://semver.org/


rob-baron (commented):
If you are going to do this, then do so for all of the docker files, not just the moc-xdmod.

Keep in mind, by doing this, we are doing something that even the xdmod team doesn't do, even though they are using docker to develop xdmod.

larsks:
> If you are going to do this, then do so for all of the docker files, not just the moc-xdmod.

I'll update the PR to include builds for the other Dockerfiles. I'll ping you when I've made the update to make sure it seems sensible.

> Keep in mind, by doing this, we are doing something that the xdmod team doesn't do, even though they are using docker to develop xdmod.

I'm not sure I follow this: we're *already* using these container images; with this PR we're just making sure things build (and building in CI rather than at deploy time).

larsks:
@rob-baron the workflow now builds images for `Dockerfile.{xdmod-openstack,moc-xdmod,moc-xdmod-edv}`.

rob-baron:
> > Keep in mind, by doing this, we are doing something that the xdmod team doesn't do, even though they are using docker to develop xdmod.
> 
> I'm not sure I follow this: we're _already_ using these container images; with this PR we're just making sure things build (and building in CI rather than at deploy time).

To my knowledge, we are _already_ using these dockerfiles to build docker images using S2I on the systems where we are running them.  We are _not_ using GitHub to build and distribute the docker images until this PR.  But, I guess I am considering this from the point of view that a dockerfile is not a docker/container image which may be irrelevant from a legal perspective.

rob-baron (approved):

naved001 (approved):

PR #157, opened 2023-04-27
Hiearchy job

rob-baron:
This is in draft as it is a work in progress.  See related issues for the breakdown of tasks.


joachimweyl:
@rob-baron would this benefit from review or is this just for your use for now?

rob-baron:
@joachimweyl yes, review away.

moc_db_helper_functions.py
```diff
@@ -0,0 +1,44 @@
+import mysql.connector
```
knikolla, line 1: I understand that for you writing SQL is easier than learning how to use SQLAlchemy, however this PR would have been about 1/3 of the size and significantly more readable if you were to use it. I strongly believe it's worth the investment in the long term to understanding how it works.

I'm happy to help in any way I can.

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+                process_record(cursor, record, hierarchy["field-of-science"])
+            case "pi":
+                process_record(cursor, record, hierarchy["pi"])
+            case "openshift-project":
+                process_record(cursor, record, hierarchy["project"])
+            case "openstack-project":
+                process_record(cursor, record, hierarchy["project"])
+                record["parent_id"] = record["id"]
+                process_record(cursor, record, hierarchy["cloud-project"])
+    return 1
```
knikolla, line 361: Is there any reason this function is returning anything at all?

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+            primary key (id, create_ts) \
+        )",
+        None,
+    )
+    cursor.execute(
+        "create sequence hierarchy_db.hierarchy_db_id_seq start with 3 increment by 1;"
+    )
+    # The sha256 field is a compsite key based on all
+    cursor.execute(
+        "create table hierarchy_db.hierarchy_keys ( \
```
knikolla, line 86: I don't see you using this table anywhere. Is it necessary?

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+    #    with open("/etc/xdmod/xdmod_init.json", "r", encoding="utf-8") as file:
+    #        config = json.load(file)["database"]
+    # except IOError:
+    #    print("Ensure the xdmod-init.json file is in /etc/xdmod/xdmod_init.json")
+
+    cnx = moc_db_helper_functions.connect_to_db(config)
+    cursor = cnx.cursor(dictionary=True)
+
+    # This is here for development
+    if moc_db_helper_functions.db_exist(cursor, "hierarchy_db"):
```
knikolla, line 426: I would strongly recommend putting this either in a separate file, or behind a specific argument when calling the scripts and not part of the main execution.

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+    if hierarchy_id:
+        add_update = False
+
+        if current_dict[hierarchy_id]["name"] != rec["name"]:
+            add_update = True
+        if current_dict[hierarchy_id]["display_name"] != rec["display_name"]:
+            add_update = True
+        if current_dict[hierarchy_id]["parent_id"] != rec["parent_id00"]:
+            add_update = True
+        if add_update:
```
knikolla, line 140: All the above loops can be combined into one conditional with `or`.

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+            l4_rec = hierarchy["cloud-project"][l5["parent_id"]]
+            l4_id = l4_rec["id"]
+            pi2project_file.write(
+                f"{hierarchy['cloud-project'][l4_id]['name']}, {l5['name']}\n"
+            )
+
+    # construct the names (rename projects)
+
+
+def find_hierarchy_id(name, dictionary):
```
knikolla, line 301: I'm confused by the existence of this function. You already use `pi_id` as the key in the dictionary, therefore you can query PI information just by doing `hierarchy["pi"][pi_id]`. Is there something I'm missing?

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+    """
+    hierarchy_id = find_hierarchy_id(rec["name"], current_dict)
+    if hierarchy_id:
+        add_update = False
+
+        if current_dict[hierarchy_id]["name"] != rec["name"]:
+            add_update = True
+        if current_dict[hierarchy_id]["display_name"] != rec["display_name"]:
+            add_update = True
+        if current_dict[hierarchy_id]["parent_id"] != rec["parent_id00"]:
```
knikolla, line 138: This seems like a typo `parent_id00`

knikolla (changes requested):

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+
+        project_rec = {
+            "name": record["attributes"]["Allocated Project ID"],
+            "display_name": record["attributes"]["Allocated Project Name"],
+            "parent_id": record["project"]["pi"],
+            "status": record["status"],
+        }
+
+        if (
+            record["resource"]["name"] == "NERC"
```
knikolla, line 394: The OpenShift resource name is `NERC-OCP`.

knikolla (changes requested):

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+            current_dict[hierarchy_id]["type"] = rec["type"]
+            current_dict[hierarchy_id]["name"] = rec["name"]
+            current_dict[hierarchy_id]["display_name"] = rec["display_name"]
+            current_dict[hierarchy_id]["status"] = rec["status"]
+            current_dict[hierarchy_id]["parent_id"] = rec["parent_id"]
+            current_dict[hierarchy_id]["still_active"] = True
+    else:
+        # create a new record with a new id
+        cursor.execute(
+            "insert into hierarchy_db.hierarchy_rec ( \
```
larsks, line 179: These two `insert` statements are broadly similar. Is there any way we can rewrite this function so that there's a single `insert` statement instead? That makes things a bit more maintainable because we won't need to make changes in two places whenever the schema changes.

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+        result = (
+            moc_db_helper_functions.exec_fetchall(
+                cursor,
+                "select id from hierarchy_rec hr where hr.type=%(type)s and hr.name=%(name)s",
+                {"type": rec["type"], "name": rec["name"]},
+            )
+        )[0]
+        hierarchy_id = None
+        if result and "id" in result:
+            hierarchy_id = result["id"]
```
larsks, line 215: You could rewrite these lines as:

```
hierarchy_id = result.get('id')
```

If I'm reading things correcly, `result` will never be empty or None because you're asking for the first item from the result of `exec_fetchall`, so either `result` will have a value or you'll get an `IndexError`.

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+    """
+    This creates a dictionary of the records returned
+
+    currently the primary key is (id, create_ts) in order to
+    track history -
+    """
+    records = moc_db_helper_functions.exec_fetchall(cursor, sql_stmt, params)
+    ret_dict = {}
+    for rec in records:
+        ret_dict[rec["id"]] = copy.deepcopy(rec)
```
larsks, line 233: You don't need `copy.deepcopy` here; you can just set:

```
ret_dict[rec["id"]] = rec
```

You're not trying to preserve the original value of `rec` for any reason.

You could write this as a dictionary comprehension:

```
return {rec["id"]: rec for rec in records}
```

process_hierarchy.py
larsks, line 361: What is this `return 1` for? There are no other return values in this function so you might as well just stick with the default return value of `None`.

larsks (changes requested):
I've made a first pass through this PR. I'm still trying to wrap my head around some of the logic, so I may have additional comments later.

I think there are some broad issues to be addressed here:

- This code needs tests.

- The code has grown big enough that we probably shouldn't be hand writing SQL statements anymore. We would benefit from an ORM with actual models and data validation.

- Much of the code (in this repository in general; not just in this PR) would probably benefit from being encapsulated in Python classes. For example, we should be able to swap the "store files in the database" abstraction with a "store files in the filesystem" abstraction without having to change any of the client code.

I'm not suggesting that all of these changes are necessary *right now*; once there is decent test coverage and the individual line comments are addressed we can go ahead and merge this PR, but I think we'll want to plan for some architecture changes in our future work.

moc_db_helper_functions.py
rob-baron: Thank you.  

Personally, I have never found ORMs to be all that useful as I have always wanted to do something special that is nearly impossible to do with the previous ORMs.  Usually ORMs do the run of the mill stuff more easily, though if you need to do something more, it becomes hideous to override and increasingly unreadable. 

Just for you, I will explore SQLAlchemy.

knikolla: I believe we should try to keep things as simple as possible. Therefore, if we're using ORM libraries in a way that makes things become hideous for us, then we should perhaps reconsider what we're doing, rather than drop the ORM library. That has held true for my years writing and reviewing software, and I've seen the same sentiment from other developers.

process_hierarchy.py
rob-baron: I created that table to hold the ids that are constructed using the digested/hashed ids.  However I am currently assigning IDs using a sequencer, so I may just use the ids that I assign as opposed to using the digest.

This will be deleted if I don't need it.

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+    # try:
+    #    with open("/etc/xdmod/xdmod_init.json", "r", encoding="utf-8") as file:
+    #        config = json.load(file)["database"]
+    # except IOError:
+    #    print("Ensure the xdmod-init.json file is in /etc/xdmod/xdmod_init.json")
+
+    cnx = moc_db_helper_functions.connect_to_db(config)
+    cursor = cnx.cursor(dictionary=True)
+
+    # This is here for development
```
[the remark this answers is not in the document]
rob-baron: I decided to go for a separate utility.  The whole point to just adding records to the database (having no updates or deletes) is to preserve history.  We will end up being able to recreate the file (not not at this iteration) by giving a date and only collecting the records up to that date.

Deleting the database on every run in the code is an initial convenance, making initial development faster.  However this has been removed.

process_hierarchy.py
rob-baron: It is just ingrained habit to check for the null result.  For the null result, the previous insert would have to fail, so ok.

process_hierarchy.py
rob-baron: deleted

moc_db_helper_functions.py
```diff
@@ -0,0 +1,44 @@
...
+def write_file_from_db(cursor, script):
+    """As a work-a-round for RWM, share config files though the database"""
+    print(f"write {script} file_share_db")
+    data = exec_fetchall(
+        cursor,
+        "select file_name, file_data from file_share_db.file where script=%s",
+        (script,),
+    )
+
+    if data and isinstance(data, list):
```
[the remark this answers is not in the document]
msdisme: I have to agree about the tests and would like to see those built out quickly as a way to help others understand the code.  

For the future, my experience as recovering software tester is that devs writing the tests earlier in the process fundamentally improves code quality.

process_hierarchy.py
rob-baron: yeah, this has outlived it's purpose.

process_hierarchy.py
rob-baron: I believe so, this was easier to write.  I will come back to this time permitting.  If time is not permitting, will add an issue.

larsks:
I made some comments yesterday about storage abstractions  and ORM, and I thought I should back those up with some examples. I spent some time this morning putting together [this repository](https://github.com/larsks/abstract-storage-example) to demonstrate a few of the points I mentioned in my review:

- The repository implements an abstract storage interface with multiple backends, including filesystem and database support.

- The database backend uses sqlalchemy, which allows it to operate against sqlite/postgres/mariadb/etc without any code changes.

- The code is thoroughly tested.

moc_db_helper_functions.py
rob-baron: Ideally they do everything that you state they do.  However, in the companies and projects where I have work with ORMs they have been massive failures.

process_hierarchy.py
rob-baron: eliminated

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+def find_hierarchy_id(name, dictionary):
+    """returns the hierarchy_id if found in the dictionary"""
+    for rec in dictionary.values():
+        if name == rec["name"]:
+            return rec["id"]
+    return None
+
+
+def get_hierarchy_from_db(cursor):
+    """This fetches the data from the database and only keeps the most recent record based on its id"""
```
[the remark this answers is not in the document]
rob-baron: I will update the comment as it is not precise enough.

moc_db_helper_functions.py
knikolla: I don't disagree with you that there are certain situations that a ORM will not help. I have found those to be the exception rather than the general case, and I don't believe this or the software that we've been building is or has been one of those exceptions.

process_hierarchy.py
```diff
@@ -0,0 +1,451 @@
...
+
+def process_expected_data(cursor, hierarchy):
+    """This fits idealized data into the hierarchy from the database
+
+    This is currently needed as we are not getting the data from
+    keycloak yet.
+    """
+    data = get_data()
+    for record in data:
+        match record["type"]:
```
knikolla, line 338: Switch statements are a new feature in Python 3.10, and this image uses Python 3.8, so this is going to throw a syntax error.

knikolla (changes requested):

process_hierarchy.py
```diff
@@ -0,0 +1,451 @@
...
+                "select id from hierarchy_rec hr where hr.type=%(type)s and hr.name=%(name)s",
+                {"type": rec["type"], "name": rec["name"]},
+            )
+        )[0]
+        hierarchy_id = None
+        if result and "id" in result:
+            hierarchy_id = result["id"]
+        if hierarchy_id:
+            current_dict[hierarchy_id] = copy.copy(rec)
+            current_dict[hierarchy_id]["id"] = id
```
knikolla, line 212: Here you are assigning the built-in Python function `id` rather than an actual value. I'm not sure what that value should be and there's an opportunity to improve the legibility of this part of the code.

knikolla (dismissed):

process_hierarchy.py
rob-baron: Ahh, there is one id that I haven't changed to 'hierarchy_id'.

process_hierarchy.py
rob-baron: Only I assign a unique id to the PIs, so the hierarchy["pi"]["id"] is not the same as hierarchy["pi"]["name"].  This makes fetching the PI information a little more complex, however, makes the overall complexity less as I can use the same function to fetch the information at any level in the hierarchy by the name (returning a unique id).  As opposed to doing something special for each layer of the hierarchy.

Additionally, by assigning a unique id, I don't have to do any cryptological digest in order to build a composite key, nor do I have to worry if the field in the xdmod is big enough.

rob-baron:
@larsks 

Thanks.

moc_db_helper_functions.py
knikolla: I have written an example of implementing this using SQLAlchemy https://gist.github.com/knikolla/604797eb2972f0f7c2692ded076a3be8

rob-baron: How about we place using SQLAlchemy as an issue as I have the impression that writing some tests a better use of tomorrow.

process_hierarchy.py
```diff
@@ -0,0 +1,478 @@
...
+        }
+    },
+    ...
+    ]
+
+    """
+    with open("../coldfront_data.json", "r", encoding="utf-8") as coldfront_file:
+        data = json.load(coldfront_file)
+        return data
+    return None
```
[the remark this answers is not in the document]
rob-baron: ok

process_hierarchy.py
rob-baron: I rewrote this to one insert statement.

moc_db_helper_functions.py
```diff
@@ -0,0 +1,44 @@
...
+        rec = data[0]
+        file = rec[0]
+        directory = os.path.dirname(file)
+        if not os.path.isdir(directory):
+            print(f"Creating Direcotry {directory}")
+            os.makedirs(directory)
```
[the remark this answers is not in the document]
rob-baron: done.

moc_db_helper_functions.py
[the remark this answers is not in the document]
rob-baron: done

rob-baron:
@larsks,

It really depends on the larger design choices that we make regarding future direction of this project.  Sure, I could write in a more object oriented style, but for most database work and much of what has been done here, I found a more procedural to be simpler to write and figure out what the requirements are.  I have been more focused on getting this up and running than on what the ultimate design should be.

1) Are we going to target a larger audience for this, or will this remain a utility that we deploy and use?
2) What is the set of services that we expect to require to deploy projects?
3) Are we going to just target OpenShift, or are we also going to support a more general form of kubernetes?
4) How prescriptive are we going to be?

I have been cutting back on how flexible the code that I write is as I am constantly reminded that nobody is asking for that flexibility, and we need to get this done soon!  Personally, I have been more concerned with the ability to add new openstack clusters without changing any code, and making used of the Kubernetes scheduler to schedule containers on nodes that are not busy.  It is this latter design choice that let me to write a quick utility to share configuration files using the database - as there wasn't any other way I had at that time.

From my perspective, where we provide flexibility really depends on how we answer broader questions.

disclaimer, I will probably edit this in the future.

rob-baron:
interesting, my local pre-commit passed:

pc
Tabs remover.............................................................Passed
trim trailing whitespace.................................................Passed
check for merge conflicts................................................Passed
fix end of files.........................................................Passed
check for added large files..............................................Passed
check for case conflicts.................................................Passed
detect private key.......................................................Passed
yamllint.................................................................Passed
black....................................................................Passed
ShellCheck v0.9.0........................................................Passed

process_hierarchy.py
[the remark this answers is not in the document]
rob-baron: The way around this is to have a more complicated database design.  One common way to accomplish this is to keep the most recent record in current table and move records from the current table into a history table.  For example, this worked well for a database that I worked with 10K - 100k elements.  With the current table, you never have to use a limit by one, you just dump the entire table and you get the table structure.  I felt that this was a little too much for a MVP as it would take too long to implement.  This is fast for the current dataset and slower for history.

Basically, what I have implemented so far is just the history table, as I see that as an MVP.  Going forward, without modifying this part, we could add a table containing the id and most recent timestamp for that ID.  With this, a single join with no need for a limit by 1 to pull the most recent record from the history table, and probably the easiest to implement with mariadb.

With a database, like oracle, you could have the current table with the same table structure as the history table.  Then add a store procedure that gets a record from the current table and insert that record into the history table upon an update to the current table.  That way, your programs could update the current table, and the history is tracked.

test/test_001.py
```diff
@@ -148,18 +146,77 @@
...
+            '"Test Project 1-ff0e78e", , "Test-Project-1 - Test Project 1-ff0e78e"\n',
+        ],
+        "pi2project.csv": [
+            '"Test Project 1-ff0e78e", "Test Project 1-ff0e78e", "nerc_openstack"\n',
+        ],
+    },
 }
 
 
 def test_000(mocker):
```
larsks, line 215: The test name should describe the purpose of the test (so that the output of a test run is informative). Names like `test_widget_created_successfully` or `test_cannot_create_widget_when_quota_succeeded` are examples of more informative names.

test/test_001.py
```diff
@@ -148,18 +146,77 @@
...
-    ds = dataset["0"]
-    mocker.patch(
-        "process_hierarchy.get_coldfront_data", return_value=ds["coldfront_data"]
-    )
-    mocker.patch(
-        "process_hierarchy.get_all_keycloak_data", return_value=ds["keycloak_data"]
-    )
     delete_hierarchy_db.delete_db()
```
larsks, line 216: It looks like you're going to be doing this for each test. Consider setting up a fixture to handle this instead.

rob-baron: will do when I add more tests.

larsks:
These tests appear to have some external dependencies:

```
$ pytest test
.
.
.
E           mysql.connector.errors.InterfaceError: 2003: Can't connect to MySQL server on 'localhost:3306' (111 Connection refused)

../../../.local/share/virtualenvs/xdmod-cntr-Sc9gjwFV/lib64/python3.11/site-packages/mysql/connector/network.py:511: InterfaceError
=========================================================== short test summary info ===========================================================
ERROR test/test_001.py::test_empty_dataset - mysql.connector.errors.InterfaceError: 2003: Can't connect to MySQL server on 'localhost:3306' (111 Connection refused)
ERROR test/test_001.py::test_2PIs_3Projects - mysql.connector.errors.InterfaceError: 2003: Can't connect to MySQL server on 'localhost:3306' (111 Connection refused)
ERROR test/test_001.py::test_removal_project_and_pi - mysql.connector.errors.InterfaceError: 2003: Can't connect to MySQL server on 'localhost:3306' (111 Connection refused)
============================================================== 3 errors in 0.20s ==============================================================
```

If the tests require a database server, they should take care of setting one up and tearing it down (or they should mock out the database interactions, or use a sqllite database for testing, etc).

rob-baron:
Ok, all tests are now in a docker container, instructions to run them are in the README.

PR #164, opened 2023-04-27
Eliminate the xdmod-init-2 container

rob-baron:
This automatically runs xdmod-setup with the parameters given in xdmod_init.json.

Unfortunately, the xdmod team has an announcement of a new version that is not currently handled by the pexpect script, So this is a quick work-a-round until we can add the message to the pexpect script.


knikolla (approved):

larsks (commented):
If we no longer have an automated process for initializing the database, then this PR should include instructions (e.g. in the README) for how to initialize the database manually.

rob-baron:
ok, will add the instructions that xdmod-setup can be run manually.

PR #166, opened 2023-04-29
Update the documentation and add a small utility

rob-baron:
I have updated the documentation to indicate that running xdmod-setup is sometimes needed, and running the file_share_through_db.py script is needed anytime the configuration is manually modified.


Dockerfile.moc-xdmod
```diff
@@ -4,28 +4,28 @@ ENV BRANCH=xdmod10.0
 # Dependencies needed by XDMoD
 RUN yum makecache && \
     yum -y install \
-        epel-release \
-        centos-release-scl-rh \
+    epel-release \
```
knikolla, line 7: The indentation changes here are unrelated and shouldn't be part of this PR.

README.md
```diff
@@ -1,6 +1,28 @@
 # xdmod-cntr
 A project to deploy XDMoD on kubernetes/OpenShift.
 
+# temporary removal of xdmod-init container.
```
knikolla, line 4: This section would be better named "Configuring XDMod" as it is pretty general.

README.md
```diff
@@ -1,6 +1,28 @@
...
+
+In the initial setup of the cluster xdmod-init will run the xdmod-setup to
+initialize the databases, create an admin account, set the organization name,
+and some other functions.  Here is the link to some documentation of xdmod-setup
+use in the configuration of xdmod (https://open.xdmod.org/10.0/configuration.html)
+
+Occasionally xdmod-setup needs to be run to change configuration paramters, for example
```
knikolla, line 11: typo: parameters

README.md
```diff
@@ -1,6 +1,28 @@
...
+adding other accounts not found in the xdmod_init.json.  This can be done manually
+by connecting with rsh the xdmod pod (on kubernetes) and running xdmod-setup from
+the command line:
+
+    oc rsh <xdmod pod name>
+    > xdmod-setup
+
+Additionally, xdmod-init will copy the configuration directory to the database
+so anytime a change in confiration is made that modifies the files in
```
knikolla, line 20: typo: configuration

README.md
```diff
@@ -1,6 +1,28 @@
...
+/etc/xdmod, these need to be saved to the database in order to share the configuration
+with the kubernetes cron jobs to do the shredding and ingesting of data.  This
+can be manually with the following script:
```
knikolla, line 23: missing "done" after manually.

knikolla (changes requested):
Since these functions are copied from `xdmod-init.py`, you should remove the duplicated code from `xdmod-init` and replace it with a call to the `file_share_through_db.py` file.

https://github.com/CCI-MOC/xdmod-cntr/blob/7bc2db4f32b92c31aaf6ac42ad98db07c6ea6144/xdmod_init.py#L562

----

Please hold off on making the above changes until you have less time pressure, and definitely not on a weekend. The PR being up is documentation enough for the time being and we can work on getting this merged in the coming weeks.

file_share_though_db.py
larsks, line 1: What is the relationship between this file and [this one](https://github.com/CCI-MOC/xdmod-cntr/pull/157/files#diff-9f518a3223fe433d85b9d3848df578359a168d2439f4f18e7cf32d42f39ddcbb) in #157?

larsks, line 1: This filename has a typo; it should be `file_shared_through_db.py`. That's why the CI check is failing.

larsks (changes requested):

file_share_though_db.py
rob-baron: The one in 157 is a proper subset of this one, I will change it to use the one in 157

README.md
rob-baron: missed this one - got it finally

file_share_though_db.py
rob-baron: Actually, the filename misspelled through as though, and so it should have been file_share_through_db.py (I had also misspelled it one in the dockerfile as file_shared_through_db.py, so I can understand the confusion).

file_share_through_db.py
```diff
@@ -0,0 +1,55 @@
...
+script that saves xdmod's configuration to a db in order to share it
+with cron jobs that are being run in separate pods (the shredder and
+the ingestor kubernetes cron jobs)
+"""
+import os
+import json
+import moc_db_helper_functions as moc_db
+
+
+def write_file_to_db(cursor, filename, script):
```
larsks, line 11: Should this also live in `moc_db_helper_functions.py`? Or is it never used anywhere else?

moc_db_helper_functions.py
```diff
@@ -0,0 +1,51 @@
...
+        cursor,
+        "select file_name, file_data from file_share_db.file where script=%s",
+        (script,),
+    )
+
+    if data and isinstance(data, list):
+        rec = data[0]
+        file = rec[0]
+        directory = os.path.dirname(file)
+        if not os.path.isdir(directory):
```
larsks, line 39: This module is not importing the `os` package, so this will result in a `NameError` exception. This is the sort of error that tools like `pylint` are trivially able to detect.

Running appropriate pre-commit checks locally before submitting your pull requests is a timesaver for everybody.

larsks (changes requested):

moc_db_helper_functions.py
rob-baron: I ran pylint on the python file that imported this expecting that pylint would also run on this file.

will be more careful in the future.

file_share_through_db.py
rob-baron: It is only used here.  When I rework xdmod-init i'll add it then.

larsks (changes requested):
I would like to see tests for the Python code introduced in this pull request.

I think `write_file_to_db` should move into `moc_db_helper_functions`, because it's clearly the counterpart to `write_file_from_db`. I would also consider renaming those so that the names aren't so similar (maybe `write_file_to_db` and `fetch_file_from_db`?).

rob-baron:
Closing this PR as I will no longer be employed by BU as of 06-SEP-2023

No need for this draft PR.

Issue #96, opened 2022-12-07
Reorganize yaml to move specific modification for the nerc in to the nerc overlay


rob-baron:
currently we have 2 files [cron-job-xdmod-openshift-prod.yaml, cron-job-xdmod-openshift-staging.yaml) that should be refactored to just cj-xdmod-openshift.yaml (for the generic file) and an associated patch file in the overlay for the nerc (to handle the deployment to the nerc).

PR #168, opened 2023-05-03
Update prod cluster URLs

tzumainn:
These parameters were never updated when the prod cluster URLs changed, leading to SSL certificate errors.

### DIFF
--- a/k8s/kube-base/cron-job-xdmod-openshift-prod.yaml
+++ b/k8s/kube-base/cron-job-xdmod-openshift-prod.yaml
@@ -31,11 +31,11 @@ spec:
               imagePullPolicy: Always
               env:
                 - name: OPENSHIFT_PROMETHEUS_URL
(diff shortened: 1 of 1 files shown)


tzumainn:
@larsks I tested these and I thiiiink these are the right URLs; would it be possible to confirm?

@rob-baron let me know what you think!

larsks (approved):
Those look correct.

rob-baron:
Sorry, I forgot to hit comment.

I kinda feel that system specific parameters should be placed into overlays so that we can support a generic instance in the kube-base and specific installations in the overlay directories.  This way we wouldn't have cron-job-xdmod-openshift-prod and cron-job-xdmod-openshift-staging but just cj-xdmod-openshift in kube_base (perhaps tailored for ocp-staging.massopen.cloud) and the specific parameters for production in k8s/overlays/nerc-ocp-infra/ and for staging in k8s/overlays/nerc-ocp-infra.  It is something that I plan to change when I find the time as I created this issue (https://github.com/CCI-MOC/xdmod-cntr/issues/96).

But this is fine for now.

Issue #170, opened 2023-05-07
configure supremm for xdmod


rob-baron:
From reading the documentation, I suspect that we need to configure supremm to get the memory figures for hpc (OpenShift).

joachimweyl:
@rob-baron what are the next steps for this issue? I belive there is communication with XDMoD on this issue can we include some of that here?

rob-baron:
@joachimweyl 

So far the dependencies for Supremm have been configured, specifically:

1. Mongo
2. Adding the xdmod usernames and passwords to mongo
3. Installing Supremm via an RPM and building the image on the system

According to the xdmod documentation (https://supremm.xdmod.org/10.0/supremm-configuration.html) there should be an Supremm option in the main menu of xdmod-setup.  Unfortunately, we don't have that on ocp-staging, so either there is a requirement that was not installed or I've missed a step or something didn't install correctly.

I've contacted the xdmid-team with all of the necessary information.  I've checked in phase 1 as it gets us closer to being able to run Supremm.

So the next step is to do what the xdmcd-team tells us to do.

rob-baron:
So far:
1) all components from the RPM were installed
2) xdmod-setup has supremm in it's main menu
3) the databases were setup (each one required additional grants for the xdmod user - similar to all of the other databases used by xdmod.  The other database grants are handled by xdmod-init.

What is left to do:
4) finish the configuration
5) figure out the next steps to get this part working

finally, deploy to the infra cluster.

rob-baron:
Have fixed the aggregate-supremm issue, so aggregate-supremm runs without generating any warning or errors.

The question has become why is no data flowing from the jobs tables to the supremm tables.  I suspect it could be a combination of:

1) Incorrect data
2) supremm configuration issue

rob-baron:
In the process of checking the data.

first of all, the table that stores the job information (hpcdb_job) has an integer field that stores the number of CPUs, then the supremm table has a floating point that stores the number of cpus.

Not sure of the math that supremm uses, but it could convert the integer from the hpcdb_job to a floating point by using a known size of a compuational node - or by the know size of the hoc cluster.  In order to test this out in staging before production, we should 

1. shred and ingest data that has milli cpu as its cpu unit.  This way 0.128 becomes an integer and can be set in the field in the jobs table.

2.  Configure supremm on staging

3.  aggregate supremm (and see if data is populated).

[...]

PR #171, opened 2023-05-08
Configure supremm - phase 1

rob-baron:
This configures the Supreme for xdmod.

#170

### DIFF
--- a/Dockerfile.moc-xdmod
+++ b/Dockerfile.moc-xdmod
@@ -65,9 +65,9 @@ RUN mkdir -p /root/rpmbuild/RPMS/noarch \
     && pip3 install keystoneauth python-keystoneclient python-novaclient python-cinderclient cryptography python-openstacksdk mysql-connector-python pexpect
 
 # RPM install of xdmod, some mongo db tools and Supremm
-RUN yum -y install  https://github.com/ubccr/xdmod/releases/download/v10.0.0/xdmod-10.0.0-1.0.el7.noarch.rpm && \
-        https://fastdl.mongodb.org/tools/db/mongodb-database-tools-rhel70-x86_64-100.7.0.rpm && \
-        https://downloads.mongodb.com/compass/mongodb-mongosh-1.8.2.x86_64.rpm && \
+RUN yum -y install  https://github.com/ubccr/xdmod/releases/download/v10.0.0/xdmod-10.0.0-1.0.el7.noarch.rpm \
+        https://fastdl.mongodb.org/tools/db/mongodb-database-tools-rhel70-x86_64-100.7.0.rpm \
+        https://downloads.mongodb.com/compass/mongodb-mongosh-1.8.2.x86_64.rpm \
         https://github.com/ubccr/xdmod-supremm/releases/download/v10.0.0/xdmod-supremm-10.0.0-1.0.el7.noarch.rpm && \
     yum clean all && \
     rm -rf /var/cache/yum


rob-baron:
I wanted this to be in draft as I am still working on it.  However I wanted to place this in a spot that Mainn could potentially find it and work on since it is impacting the job side of xdmod.

tzumainn:
@rob-baron thanks for sharing! I assume that the new pvcs would have to be added to the xdmod jobs?

rob-baron:
no, this is a separate pv just for the mongo database, like the pv for mariadb.  So mongo is a service that the ingestor will do some work on.  It doesn't involve the shredder and the PV should be completely under mongo's control.

rob-baron:
This add supremm and some mongo tools to the xdmod container as well ass adding a mongo container in a similar manner to mariadb.

haven't worked out the full configuration of supremm, however this get a significant way there while the xdmod team gets back to me concerning this.  So consider this phase 1.

Dockerfile.moc-xdmod
```diff
@@ -64,8 +64,11 @@ RUN mkdir -p /root/rpmbuild/RPMS/noarch \
...
     && pip3 install keystoneauth python-keystoneclient python-novaclient python-cinderclient cryptography python-openstacksdk mysql-connector-python pexpect
 
-# RPM install of xdmod
+# RPM install of xdmod, some mango db tools and Supremm
 RUN yum -y install  https://github.com/ubccr/xdmod/releases/download/v10.0.0/xdmod-10.0.0-1.0.el7.noarch.rpm && \
+    yum -y install  https://fastdl.mongodb.org/tools/db/mongodb-database-tools-rhel70-x86_64-100.7.0.rpm && \
+    yum -y install  https://downloads.mongodb.com/compass/mongodb-mongosh-1.8.2.x86_64.rpm && \
+    yum -y install  https://github.com/ubccr/xdmod-supremm/releases/download/v10.0.0/xdmod-supremm-10.0.0-1.0.el7.noarch.rpm && \
     yum clean all && \
     rm -rf /var/cache/yum
```
larsks, line 73: It's generally faster to use a single `yum` command rather than multiple `yum` commands (because that way you only need to calculate dependencies once):

```suggestion
RUN yum -y install \
  https://github.com/ubccr/xdmod/releases/download/v10.0.0/xdmod-10.0.0-1.0.el7.noarch.rpm \
  https://fastdl.mongodb.org/tools/db/mongodb-database-tools-rhel70-x86_64-100.7.0.rpm \
  https://downloads.mongodb.com/compass/mongodb-mongosh-1.8.2.x86_64.rpm \
  https://github.com/ubccr/xdmod-supremm/releases/download/v10.0.0/xdmod-supremm-10.0.0-1.0.el7.noarch.rpm \
  && \
  yum clean all && \
  rm -rf /var/cache/yum
```

k8s/kube-base/deployment-mongo.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  creationTimestamp: null
```
larsks, line 4: You can drop `creationTimestamp` from the manifest.

Dockerfile.moc-xdmod
```diff
@@ -64,8 +64,11 @@ RUN mkdir -p /root/rpmbuild/RPMS/noarch \
     && pip3 install --upgrade pip \
     && pip3 install keystoneauth python-keystoneclient python-novaclient python-cinderclient cryptography python-openstacksdk mysql-connector-python pexpect
 
-# RPM install of xdmod
+# RPM install of xdmod, some mango db tools and Supremm
```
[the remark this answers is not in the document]
rob-baron: yup, spell check really wants to go from mongo to mango.

will change.

larsks (approved):

PR #173, opened 2023-05-10
Change the unknow flavor's cpu, memory and disk values to 1

rob-baron:
Xdmod cannot handle 0 for cpu. I have made mem and disk also 1 as I suspecit there are places where having 0 for these values may cause failure.


joachimweyl (approved):
0 to 1 seems like a pretty simple change.

PR #182, opened 2023-06-05
Moved time of xdmod openshift cron job

tzumainn:
The job currently runs at 1am; however 1am server time is apparently 9pm the previous day OpenShift time. Changing the time to 1pm will allow the job to get all the data for the previous day.

### DIFF
--- a/k8s/kube-base/cron-job-xdmod-ingestor.yaml
+++ b/k8s/kube-base/cron-job-xdmod-ingestor.yaml
@@ -3,7 +3,7 @@ kind: CronJob
 metadata:
   name: xdmod-ingestor
 spec:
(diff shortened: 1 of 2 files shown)


tzumainn:
@rob-baron Let me know what you think! If this is okay, after this is merged can you also push the change live?

rob-baron (changes requested):
Lets also move the time that the ingestor runs to around 2:00, as this PR implies that.

tzumainn:
> Lets also move the time that the ingestor runs to around 2:00, as this PR implies that.

Sounds good! I've updated the time.

rob-baron (approved):
Look good.

PR #187, opened 2023-06-14
Fixes the manifests for mongo

rob-baron:
1) place the reference to the mongo pic in kustomization
2) increase the size of the pv for mongo


larsks (approved):

rob-baron:
@kim - this is just an order of magnitude.  I am hoping that this will be enough so that I can get it up and running and see how quickly mongo grows.

